PR #3, opened 2022-10-23
feat(devx): updated dev script to work on Windows machines

lxkedinh:
- Improves on the dev script by @exelarios to allow Windows developers to run the entire stack and set up ngrok with a single command
- Developers must have the folder containing their ngrok executable set in their Path Windows system environment variable.

### DIFF
--- a/scripts/run.js
+++ b/scripts/run.js
@@ -1,27 +1,67 @@
 const fs = require("fs");
 const os = require("os");
 const path = require("path");
-const { exec, execSync } = require("child_process");
+const { exec, execSync, spawn } = require("child_process");
 
 const PORT = 5050;
+const PATH_LIMIT = 1024;
+const SPAWN_SHELL_DELAY = 1000;
+const REWRITE_NGROK_URL_DELAY = 3000;
+const NGROK_PATH = process.argv[2];
 const systemType = os.platform();
 
-if (systemType !== "darwin") {
-  console.log("This command currently only supports unix at the moment.");
-  return;
+const windowsShellOptions = {
+  shell: process.env.ComSpec || "C:\\Windows\\system32\\cmd.exe",
+};
+
+function setNgrokPath() {
+  if (!NGROK_PATH) return;
+
+  const newPath = process.env.path + ";" + NGROK_PATH;
+
+  if (newPath.length > PATH_LIMIT) {
+    throw new Error(
+      "Your path variable is over 1024 characters long. You will have to shorten it or add ngrok manually to your path."
+    );
+  } else {
+    console.log("path changed");
+    exec(`setx PATH "%PATH%;${NGROK_PATH}"`);
+  }
 }
 
 function newTerminal(command) {
-  const PWD = path.join(__dirname, "../");
-  execSync(`
-  osascript -e 'tell application "Terminal" to activate' \
-  -e 'tell application "System Events" to keystroke "t" using {command down}' \
-  -e 'tell application "Terminal" to do script "cd ${PWD} && ${command}" in front window'
-  `);
+  const CWD = path.join(__dirname, "../");
+
+  switch (systemType) {
+    case "win32":
+      // empty double quotes is intentional because of how the "start" Windows
+      // command parameters work, setTimeout() used for Windows to prevent new
+      // shells from opening too fast and skipping some commands
+      setTimeout(() => {
+        spawn(`start "" /d ${CWD} ${command}`, [], windowsShellOptions);
+      }, SPAWN_SHELL_DELAY);
+      break;
+    case "darwin":
+      execSync(`
+      osascript -e 'tell application "Terminal" to activate' \
+      -e 'tell application "System Events" to keystroke "t" using {command down}' \
+      -e 'tell application "Terminal" to do script "cd ${CWD} && ${command}" in front window'
+      `);
+      break;
+    default:
+      console.log(
+        "This dev script currently only supports Windows/unix at the moment."
+      );
+  }
 }
 
 // Make sure all instances of ngrok's process doesn't exist.
-exec("killall ngrok");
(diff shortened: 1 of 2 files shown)


scripts/run.js
```diff
@@ -6,22 +6,30 @@ const { exec, execSync } = require("child_process");
...
-  -e 'tell application "Terminal" to do script "cd ${PWD} && ${command}" in front window'
-  `);
+  const CWD = path.join(__dirname, "../");
+
+  // empty double quotes is intentional because of how the "start" Windows
+  // command parameters work
+  if (systemType === "win32") {
+    setTimeout(() => {
+      exec(`start "" /d ${CWD} ${command}`);
+    }, 1000);
```
exelarios: Why is `setTimeout` needed here?

scripts/run.js
```diff
@@ -6,22 +6,30 @@ const { exec, execSync } = require("child_process");
...
+  } else
+    console.log(
+      "This dev script currently only supports Windows/unix at the moment."
+    );
 }
 
 // Make sure all instances of ngrok's process doesn't exist.
-exec("killall ngrok");
+if (systemType === "win32") exec('Taskkill /IM "ngrok.exe" /F');
+else if (systemType === "darwin") exec("killall ngrok");
```
exelarios: Please have curly braces for all statements, even if it's not required.
Also, please have the next statement on the next line.

This will allow better readability and keeps the codebase consistent.

scripts/run.js
```diff
@@ -6,22 +6,30 @@ const { exec, execSync } = require("child_process");
...
-  execSync(`
-  osascript -e 'tell application "Terminal" to activate' \
-  -e 'tell application "System Events" to keystroke "t" using {command down}' \
-  -e 'tell application "Terminal" to do script "cd ${PWD} && ${command}" in front window'
-  `);
+  const CWD = path.join(__dirname, "../");
+
+  // empty double quotes is intentional because of how the "start" Windows
+  // command parameters work
+  if (systemType === "win32") {
```
exelarios: Can you convert this condition block into a switch statement?

exelarios (changes requested):
Tested on macOS, I can confirm the command still works on MacOS.

We just need one more reviewer from a Windows user to verify if the command works.

Overall, nice job! Looks good for the most part.

exelarios:
> Developers must have the folder containing their ngrok executable set in their Path Windows system environment variable.

Can you see if is it possible to add ngrok automatically into their path if they are running the command for the first time?

scripts/run.js
lxkedinh: I was running into a problem while testing where having just the `exec()` statements would open terminals too fast and skip the commands in a sense. The `ngrok HTTP PORT` command would not be executed while the new terminal windows for `yarn app:start` and `yarn server:start` opened fine. Adding a one second delay with `setTimeout()` fixed that problem for me.

lxkedinh:
> > Developers must have the folder containing their ngrok executable set in their Path Windows system environment variable.
> 
> Can you see if is it possible to add ngrok automatically into their path if they are running the command for the first time?

I have found a way to programmatically add ngrok to a developer's path variable using [setx](https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/setx#additional-references), but the developer will have to add it manually if their current path variable plus the ngrok path is over 1024 characters as the command automatically crops text over this limit, which will result in loss of data to some other programs in their path variable.

image

scripts/run.js
lxkedinh: resolved on commit [bb7b3e2](https://github.com/cppsea/icebreak/pull/3/commits/bb7b3e2ed0c5fec240287e223167e51dc985351a)

scripts/run.js
lxkedinh: resolved on commit [bb7b3e2](https://github.com/cppsea/icebreak/pull/3/commits/bb7b3e2ed0c5fec240287e223167e51dc985351a)

exelarios:
@lxkedinh 
Awesome find!
If we can get the path, we could check if the path is over 1024 characters and throw an error that they must manually set their own `ngrok` path.

exelarios (approved):

GuyWhoCode (approved):
The changes in the pull request work on Windows.

lxkedinh:
@exelarios  I implemented automatic ngrok path environment variable setting with error handling for going over the character limit. Can you take a look at commit [f990ae8](https://github.com/cppsea/icebreak/pull/3/commits/f990ae82546e08744b6bee479d5d83b1b82472fc)?

scripts/run.js
```diff
@@ -1,27 +1,72 @@
...
 
-if (systemType !== "darwin") {
-  console.log("This command currently only supports unix at the moment.");
-  return;
-}
+const windowsShellOptions = {
+  shell: process.env.ComSpec || "C:\\Windows\\system32\\cmd.exe",
+};
+
+const setNgrokPath = () => {
```
exelarios: I know it doesn't matter in this case, but please use a normal function instead of an arrow function, especially if it's in the global scope unless the function is passed as a parameter or within a function.

Sorry, I'll write up a coding convention for this repo later.

exelarios (approved):
It still works on MacOS. So everything is fine by me.

scripts/run.js
```diff
@@ -1,27 +1,72 @@
...
 
 // Make sure all instances of ngrok's process doesn't exist.
-exec("killall ngrok");
+if (systemType === "win32") {
+  try {
+    setNgrokPath();
+  } catch (err) {
+    console.error(err);
+    throw err;
+  }
```
exelarios: What's the point of catching the error, if you're going to throw the error anyways? It defeats the purpose of having a `try..catch` block. Just remove the `try..catch` block altogether, since it doesn't matter if the script crashes.

PR #4, opened 2022-10-31
feat(database): add guild & events schema into the database

exelarios:
- Implement Events and Guilds schema into the database.
- Inserted dummy data into both events and guilds.

### DIFF
--- a/packages/server/controllers/events.js
+++ b/packages/server/controllers/events.js
@@ -1,0 +1,16 @@
+const postgres = require("../utils/postgres");
+
+async function getAllEvents() {
+  const query = await postgres.query('SELECT * FROM event');
+  return query.rows;
+}
+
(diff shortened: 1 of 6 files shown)


E-A-H62 (approved):
Everything looks good, and I think I understood the basics of each file.  Although, I'm just a bit confused with the files in routes/api.  Are those used to test something?

exelarios:
@E-A-H62 Not really to test, put to declare the actual route path explicitly. So like, if the user goes to `http://localhost:5050/api/events`, it would use the `api/events` file, and inside the file, it has nested routes that belong to the parent route.

The controller's directory that uses holds the majority of the logic for the routes. The controller file name should corresponding to the route file name.

lxkedinh (approved):

PR #6, opened 2022-11-14
chore(db): create event schemas

GuyWhoCode:
- Added schemas not already implemented 
- Double checked existing schemas
Note: These are the schemas noted in the Technical Design Document. More schemas might be added later but will be added when needed.

### DIFF
--- a/packages/server/scripts/database.sql
+++ b/packages/server/scripts/database.sql
@@ -6,7 +6,7 @@ CREATE TABLE users (
   last_name varchar(50) NOT NULL,
   email varchar(255) UNIQUE NOT NULL,
   avatar varchar(255) NOT NULL,
-  password TEXT
+  password varchar(255)
 );
 
 CREATE TABLE Guild (
--- a/packages/server/scripts/database.sql
+++ b/packages/server/scripts/database.sql
@@ -38,6 +38,9 @@ CREATE TABLE Event (
 CREATE TABLE user_guild (
   user_id VARCHAR(255),
   guild_id VARCHAR(255),
+  user_role VARCHAR(255),
+  points SMALLINT,
+  rank SMALLINT
   FOREIGN KEY(user_id)
     REFERENCES users(user_id)
     ON UPDATE CASCADE
@@ -47,9 +50,6 @@ CREATE TABLE user_guild (
     ON UPDATE CASCADE
     ON DELETE SET NULL,
   PRIMARY KEY(user_id, guild_id),
-  user_role VARCHAR(255),
-  points SMALLINT,
-  user_admin SMALLINT
 );
 
 CREATE TABLE members_pending (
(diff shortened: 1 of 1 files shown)


packages/server/scripts/database.sql
```diff
@@ -34,6 +35,36 @@ CREATE TABLE Event (
...
+    ON UPDATE CASCADE
+    ON DELETE SET NULL,
+  FOREIGN KEY(guild_id)
+    REFERENCES Guild(guild_id)
+    ON UPDATE CASCADE
+    ON DELETE SET NULL,
+  PRIMARY KEY(user_id, guild_id),
+  user_role VARCHAR(255),
+  points SMALLINT,
+  user_admin SMALLINT
```
exelarios: Can you move these above the keys function please?

packages/server/scripts/database.sql
```diff
@@ -1,18 +1,19 @@
...
+  avatar varchar(255) NOT NULL,
+  password varchar(255)
 );
 
 CREATE TABLE Guild (
   guild_id VARCHAR(255),
   name VARCHAR(100),
   handler VARCHAR(50),
-  description VARCHAR(255),
+  description TEXT,
```
exelarios, line 16: What's is the difference of using `TEXT` compared to `VARCHAR` here?
I think we should at least have a limit of the amount of characters they can have? Maybe limit it to 500 characters?

exelarios (changes requested):
Great job on making the changes and investigating into our database schema.

Once the changes are addressed, feel free to merge it into main.

packages/server/scripts/database.sql
exelarios: I think it's best to just name this `rank` instead. The higher the `rank` value the more privilege access to the Guild. For instance, a Guild member can post an event if they are above `100`.

packages/server/scripts/database.sql
GuyWhoCode: From this StackOverflow article (https://stackoverflow.com/a/25301046), the memory of TEXT and VARCHAR will be the same since the expected amount of characters used will be beyond 256 characters. It would be nice to limit the amount of characters used but since the memory usage is the same, it's better in my opinion to account for the possibility of clubs going over the 500 character limit.

packages/server/scripts/database.sql
GuyWhoCode: Can you clarify what column of this table to be renamed to rank? You selected 3 possible columns in the table in your comment.

GuyWhoCode: Can you clarify what column of this table needs to be renamed to rank? You selected 3 possible columns in the table in your comment.

exelarios: I was thinking of the `user_admin`.

exelarios (approved):
Looks good. Feel free to merge it if the changes are also migrated into the database.

marked01one (approved):
Asked @GuyWhoCode  to try running edge cases where there are repeated `user_id` and `guild_id` elements in `user_guild` and he reported that the elements were accepted. 

Approved by @marked01one

PR #7, opened 2022-11-24
feat(events): populating events on the feed 

E-A-H62:
Created refresh handling for the feed screen and populated feed screen with event card component(s) when the user logs in.


exelarios (approved):
I tested out the refresh feed feature by adding a new event to the database, looks like it works as expected.
RPReplay_Final1669275748_MP4_AdobeExpress

Great job team!

GuyWhoCode (approved):

PR #8, opened 2022-12-02
feat(auth): implement local auth

JustinNgxyen:
**Changes**: updated regex algorithm for register. 

**Testing**: Postman via POST request (http://localhost:5050/api/auth/login)


exelarios:
Please stick to only using `yarn` package manager. If you don't, it will generate a `package-lock.json` which could cause conflicts with the `yarn-lock.json`. Please remove the entire `package-lock.json` from your PR. This PR shouldn't have more than 10k lines of changes. Plus, it seems to be something odd about the `yarn-lock.json`. I don't understand why there are more than 10k changes in `yarn-lock.json`.

Another thing about this PR is, how this tested?

packages/server/routes/api/auth.js
```diff
@@ -2,11 +2,18 @@ const express = require("express");
 const passport = require("passport");
 const router = express.Router();
 const token = require("../../utils/token");
-const { OAuth2Client } = require('google-auth-library');
+const { OAuth2Client, UserRefreshClient } = require('google-auth-library');
```
exelarios, line 5: What is `UserRefreshClient` being used for?

packages/server/routes/api/auth.js
```diff
@@ -2,11 +2,18 @@ const express = require("express");
...
 const GoogleStrategy = require("passport-google-oauth20").Strategy;
 
 const AuthController = require("../../controllers/auth");
+const { response } = require("express");
+const postgres = require("../../utils/postgres");
+const user = require("../../controllers/users");
+
+//const crypto = require('node:crypto');
+const bcrypt = require('bcrypt');
+const uniqid = require('uniqid');
```
exelarios, line 16: Please have all your imports on top on the file. To keep things consistent.

packages/server/routes/api/auth.js
```diff
@@ -107,4 +114,86 @@ router.get("/google/callback", passport.authenticate("google", {
...
-module.exports = router;
\ No newline at end of file
+router.post("/register", async (request, response) => {
+
+  try{
+    const { email, password, } = request.body;
+    
+    if(!email.includes("@") || email.includes(" ")){ // check if email is valid
+      throw new Error("Email is invalid.");
+    }
```
exelarios, line 124: Please use regex to validate email.

exelarios (changes requested):

packages/server/routes/api/auth.js
```diff
@@ -2,11 +2,18 @@ const express = require("express");
 const passport = require("passport");
 const router = express.Router();
 const token = require("../../utils/token");
-const { OAuth2Client } = require('google-auth-library');
+const { OAuth2Client, UserRefreshClient } = require('google-auth-library');
 
 const GoogleStrategy = require("passport-google-oauth20").Strategy;
 
 const AuthController = require("../../controllers/auth");
+const { response } = require("express");
```
GuyWhoCode, line 10: This might be the result of VSCode autocomplete. You should remove it since it's an unnecessary import.

packages/server/routes/api/auth.js
```diff
@@ -107,4 +114,86 @@ router.get("/google/callback", passport.authenticate("google", {
...
+  try{
+    const { email, password, } = request.body;
+    
+    if(!email.includes("@") || email.includes(" ")){ // check if email is valid
+      throw new Error("Email is invalid.");
+    }
+
+    const ifExist = await user.getUserByEmail(email);
+
+    if(ifExist?.email == email){ // check if email is already in the database
```
GuyWhoCode, line 128: Extremely nitpicky and you don't have to change it, but you should use === to check for both type and value.

GuyWhoCode (approved):
You should implement deric's suggestions before merging. Otherwise, great progress!

packages/server/routes/api/auth.js
exelarios: No, I highly recommend to change it to reduce unexpected bugs with types. It's always best to use strictly equally whenever you can.

packages/server/routes/api/auth.js
```diff
@@ -107,4 +114,89 @@ router.get("/google/callback", passport.authenticate("google", {
...
 
-module.exports = router;
\ No newline at end of file
+router.post("/register", async (request, response) => {
+
+  try{
+    const { email, password, } = request.body;
+    
+    let emailCheck = /@/;
+    let whitespaceCheck = /\s/;
```
exelarios: The regex here is valid, but not strong enough. I recommend using something like this:
https://regexr.com/3e48o

If your regex is strong enough there's no need to test for an `@` and whitespace separately.

exelarios (changes requested):

packages/server/routes/api/auth.js
```diff
@@ -107,4 +114,88 @@ router.get("/google/callback", passport.authenticate("google", {
...
+
+    bcrypt.genSalt(saltRounds, function(err, salt) {
+      bcrypt.hash(password, salt, async function(err, hash) { // encrypt the user password using bcrypt
+
+        // create unique User ID as bytes (18 byte)
+        const user_id = uniqid();
+
+        await postgres.query(`
+          INSERT INTO users (user_id, first_name, last_name, email, avatar, password)
+          VALUES ('${user_id}', 'firstName', 'lastName', '${email}', 'avatar', '${password}');
```
exelarios: Why are you storing the plain-text password into the database? Shouldn't you be storing the encrypted password instead?

packages/server/routes/api/auth.js
```diff
@@ -107,4 +114,88 @@ router.get("/google/callback", passport.authenticate("google", {
...
+        `); // create new user in DB
+
+        const newToken = token.generate({ email, hash}); // create a signed jwt token
```
exelarios: Do not store sensitive data inside the JWT token, it's not an encryption.

The payload being passed into the token is in plain text and can be seen by anyone.
https://jwt.io/

packages/server/routes/api/auth.js
```diff
@@ -107,4 +114,88 @@ router.get("/google/callback", passport.authenticate("google", {
...
+  try{
+    
+    // Get user input
+    const { email, password } =  request.body;
+    const ifExist = await user.getUserByEmail(email);
+    const hashedPassword = ifExist.password;
+
+    // Validate credentials
+    if(!(email && password)){
+      response.status(403).send("All inputs are required");
```
exelarios, line 177: When the consuming the API expected response in JSON, it should be all in JSON format including the data or it could cause unexpected errors on the client-side.

packages/server/routes/api/auth.js
```diff
@@ -107,4 +114,88 @@ router.get("/google/callback", passport.authenticate("google", {
...
+    }
+
+    // Validate if email is in database
+    if(user.getUserByEmail(email) && await bcrypt.compare(password, hashedPassword)){ // check if email is already in the database
+        // Create token
+        const newToken = token.generate({ email, hashedPassword});
```
exelarios, line 183: ditto: JWT token

packages/server/routes/api/auth.js
```diff
@@ -107,4 +114,88 @@ router.get("/google/callback", passport.authenticate("google", {
...
+        response.send({  // send jwt token
+          success: true,
+          newToken
+          }
+        ); 
+      }else{
+      response.status(403).send("Wrong email and/or password.");
```
exelarios, line 190: ditto: JSON format

exelarios (changes requested):
Please remove `package-lock.json` from this PR, since we're using `yarn`.

I tested out the register and login endpoints. The login endpoint doesn't seem to work. I suspect it's because it's comparing the incorrect password.

I'm a bit disappointed in this PR. There are many critical issues that can easily be discovered if the endpoints were at least tested through Postman. Please do not request for review without being fully confident that your feature works.

exelarios (approved):
Although there were many issues with the indentations and potentially could have better variable naming. It's also missing the implementation of refresh tokens, which would be a implemented follow-up sprint.

Everything else, LGTM. Push the code whenever y'all feel ready.

PR #10, opened 2023-01-10
feat(event): event card component

marked01one:
The `EventCard` component along with its sub-components are located inside the `@app/components/EventCard` directory.


packages/app/components/EventCard/EventCard.js
```diff
@@ -0,0 +1,51 @@
...
+import { View, Image, Text, FlatList, StyleSheet } from 'react-native';
+import { Card, ListItem, Icon, Divider } from 'react-native-elements';
+import EventCardText from './EventCardText';
+import EventCardRegistration from './EventCardRegistration';
+
+
+
+const styles = StyleSheet.create({
+  banner: {
+    width: 360,
```
GuyWhoCode: I recommend looking into relative CSS units like `vw` in this case instead of pixel values. See screenshot of the empty space in the frontend channel.

packages/app/components/EventCard/EventCard.js
```diff
@@ -0,0 +1,51 @@
+import React, { useCallback, useEffect, useState} from 'react';
```
GuyWhoCode: Your code editor will highlight which imported variables are not being used. I suggest you remove the ones you're not using.

packages/app/components/EventCard/EventCardRegistration.js
```diff
@@ -0,0 +1,52 @@
...
+}); 
+
+export default function EventCardRegistration(props) {
+  const { register } = props;
+  
+  return (
+    <View style={{ flexDirection: 'row' }}>
+      <View style={styles.buttonView}>
+        <Button
+          icon={<Icon color='#3498db' />}
```
GuyWhoCode: Is there some problem that forces you to change the button color using an icon instead of with CSS?

GuyWhoCode (changes requested):
See comments listed.

packages/app/components/EventCard/EventCardRegistration.js
marked01one: I only added the color code in there to make it clear for anyone who reads it. It doesn't serve any functional purposes. If you deem that it is unnecessary, I'll remove it.

GuyWhoCode: You are already adding CSS styling to the button, so to me, it feels redundant to explicitly add a color code there as its own property instead of being part of the existing button styling.

packages/app/components/EventCard/EventCard.js
```diff
@@ -1,14 +1,14 @@
...
+import { Card, Divider } from 'react-native-elements';
 import EventCardText from './EventCardText';
 import EventCardRegistration from './EventCardRegistration';
 
 
 
 const styles = StyleSheet.create({
   banner: {
-    width: 360,
+    width: '100%',
```
marked01one: Changed `width` property from using pixels to percentages

packages/app/components/EventCard/EventCard.js
```diff
@@ -1,14 +1,14 @@
-import React, { useCallback, useEffect, useState} from 'react';
-import { View, Image, Text, FlatList, StyleSheet } from 'react-native';
-import { Card, ListItem, Icon, Divider } from 'react-native-elements';
+import React from 'react';
```
marked01one: Removed unnecessary imports

packages/app/components/EventCard/EventCard.js
exelarios: Percentage would be the correct unit, since `vw` would reference the screen size rather than the parent container's width.

packages/app/components/EventCard/EventCard.js
```diff
@@ -0,0 +1,51 @@
...
+import { Image, StyleSheet, Dimensions } from 'react-native';
+import { Card, Divider } from 'react-native-elements';
```
exelarios: Please create your own components. Please do not use react-native-elements or any third-party components libraries.

packages/app/components/EventCard/EventCardRegistration.js
```diff
@@ -0,0 +1,52 @@
+import React from "react";
+import { StyleSheet, Button, View } from "react-native";
+import { Icon } from "react-native-elements";
```
exelarios: ditto: use react-native components only

packages/app/package.json
```diff
@@ -16,12 +16,13 @@
...
     "axios": "^0.27.2",
     "expo": "~46.0.13",
-    "expo-auth-session": "~3.7.1",
+    "expo-auth-session": "~3.7.4",
     "expo-random": "~12.3.0",
     "expo-status-bar": "~1.4.0",
     "expo-web-browser": "~11.0.0",
     "react": "18.0.0",
     "react-native": "0.69.6",
+    "react-native-elements": "^3.4.2",
```
exelarios: Remove this library from this project.

packages/app/screens/landing/LandingScreen.js
```diff
@@ -60,6 +61,13 @@ function LandingScreen() {
         onPress={handleOnLoginWithGoogle}
         title="login with google"
       />
+      <EventCard 
+        eventBanner={require("@app/components/EventCard/eventcard_test/test_card_banner.png")}
+        timeBegin="11:00AM" timeEnd="12:00PM"
+        eventTitle="College Survival Guide"
```
exelarios: Instead of being `eventTitle`, just name the prop name as `title`.

packages/app/screens/landing/LandingScreen.js
```diff
@@ -60,6 +61,13 @@ function LandingScreen() {
         onPress={handleOnLoginWithGoogle}
         title="login with google"
       />
+      <EventCard 
+        eventBanner={require("@app/components/EventCard/eventcard_test/test_card_banner.png")}
```
exelarios: Remove `event` from `eventBanner`.

exelarios (changes requested):
The component looks good, although stick to using react-native components and not any third-party components.

Also, please replace your face with a placeholder avatar picture. Something like this: https://upload.wikimedia.org/wikipedia/commons/7/7c/Profile_avatar_placeholder_large.png

Move all the `StyleSheet.create..` below the component declaration.

marked01one:
@GuyWhoCode please review recent changes and approve or request more changes as necessary

GuyWhoCode (approved):
Don't see any issues with your code and it works great when I tested it. Nice job!

exelarios (approved):

PR #13, opened 2023-02-01
feat(login): create login and signup form

nappalion:
Implemented the UI for the login and signup form with input for email and password.

**Current Version Demo:** https://imgur.com/fNGQsvI

**RFC:** [https://docs.google.com/document/d/1ixDwuFMW4EJTEnfl-CG6AodOlBydy4RODPEGWv7mMA4/edit?usp=sharing](url)

Changes Made:
- Added "react-native-svg": "12.3.0" to "package.json". (had to run "expo install react-native-svg" to get the working version)
- Added three SVG icons (password visibility and google)
- Created TextInput.js in "components"
- Modified LandingScreen.js in "screens" and Button.js in "components"
- Added unit tests in "Login.test.js" file


packages/app/screens/landing/LandingScreen.js
```diff
@@ -1,67 +1,296 @@
...
-      const result = await promptAsync();
+  // skips the code during test
+  if (process.env.NODE_ENV !== 'test') {
+    const [request, response, promptAsync] = Google.useAuthRequest({
+      responseType: 'id_token',
+      expoClientId: '1080245081969-u9lnl9ospj757rq75kiumttqconhnfcc.apps.googleusercontent.com',
+      iosClientId: '1080245081969-u9lnl9ospj757rq75kiumttqconhnfcc.apps.googleusercontent.com',
+      androidClientId: '1080245081969-u9lnl9ospj757rq75kiumttqconhnfcc.apps.googleusercontent.com',
+      webClientId: '1080245081969-u9lnl9ospj757rq75kiumttqconhnfcc.apps.googleusercontent.com',
+    });
```
lxkedinh: Are these supposed to be secret credentials? If they are, they should be changed asap if possible and then put in environment variables for the future since the repo's public.

lxkedinh (commented):
Overall, everything looks good to me! Screen looks really nice and runs perfectly on my Galaxy S9+.

packages/app/screens/landing/LandingScreen.js
exelarios: Yeah, we should move these into secrets. I believe I was in a rush to migrate from GoogleOAuth into using ExpoAuth Providers.

nappalion: Ok, I'll try and move them to secrets, and I'll fix the yarn/npm problem soon.

packages/app/screens/landing/SignUpScreen.js
```diff
@@ -0,0 +1,332 @@
...
+      if (response?.data.status == "success") {
+        navigation.navigate("LandingScreen", { email: inputs.email });
+      }
+    } catch (error) {
+      console.log(error.toString());
+    }
+  };
+
+  // State to change the variable with the TextInput
+  const [inputs, setInputs] = React.useState({
```
lxkedinh: You're only using the global `React` variable for `useState` in this file, so there's no point in importing it. You can just import useState itself along with the other imports and change your states to just `useState()` instead of `React.useState()`.
```javascript
import { useState, useCallback, useRef } from "react";
```
```javascript
const [inputs, setInputs] = useState({
```

packages/app/screens/landing/SignUpScreen.js
```diff
@@ -0,0 +1,332 @@
...
+    passwordConfirmation: "",
+  });
+
+  const [errors, setErrors] = React.useState({});
+
+  const isValidEmail = (email) => {
+    const emailRE =
+      /^(([^<>()[\]\.,;:\s@\"]+(\.[^<>()[\]\.,;:\s@\"]+)*)|(\".+\"))@(([^<>()[\]\.,;:\s@\"]+\.)+[^<>()[\]\.,;:\s@\"]{2,})$/i;
+    return email.match(emailRE);
+  };
```
lxkedinh: When you declare functions and variables inside React components, you need to think about whether they interact with state or not. This `isValidEmail` utility function doesn't interact with the state of this component (the user's inputs, possible error states). You should put this outside the React component (preferably near the bottom of the file) so this function isn't redefined every time the component rerenders.

packages/app/screens/landing/SignUpScreen.js
```diff
@@ -0,0 +1,332 @@
...
+import EyeOn from "@app/assets/eye-line-on";
+import * as SecureStore from "@app/utils/SecureStore";
+
+import Constants from "expo-constants";
+import DividerWithText from "@app/components/DividerWithText";
+
+WebBrowser.maybeCompleteAuthSession();
+
+function SignUpScreen({ navigation, route }) {
+  const { user, setUser } = useUserContext();
```
lxkedinh, line 36: Why do you get the user context here for the signup screen? It should be null at this point since the user is unauthenticated and registering for our app.

packages/app/screens/landing/SignUpScreen.js
```diff
@@ -0,0 +1,332 @@
...
+        password: inputs.password,
+      })
+
+      if (response?.data.status == "success") {
+        navigation.navigate("LandingScreen", { email: inputs.email });
+      }
+    } catch (error) {
+      console.log(error.toString());
+    }
+  };
```
lxkedinh: The logic in your register function is good here, but you are missing the case where the server finds an existing user with the same email and returns a fail type response.
```JSON
{
  "status": "fail",
  "data": {
    "email": "A user with this email already exists.",
  },
}
```

lxkedinh: We will most likely be implementing this with an alert dialog to show the user when they input an email that is tied to an existing user. I'll have to talk to Napoleon about possibly adding a third-party React Native modal popup/alert dialog library. For now, you can handle this case with a console.log just to make sure it works.

lxkedinh (changes requested):
Overall really impressive work, Ethan! You have a nice coding style and the signup screen looks and works really nicely so far. There are just a few things we need to fix, but great job so far.

packages/app/screens/landing/SignUpScreen.js
lxkedinh, line 1: Remove all code regarding the new user's first and last name since we're moving that to the onboarding screen.

packages/app/screens/landing/SignUpScreen.js
```diff
@@ -0,0 +1,335 @@
...
+            textStyle={styles.boldText}
+          />
+
+          <DividerWithText title="or" />
+
+          <Button
+            testID="googleButton"
+            title="Continue with Google"
+            underlayColor="#ebebeb"
+            onPress={() => {}} //TODO: Implement Google Sign Up
```
lxkedinh: Signing up with Google is the same as signing in with Google for our case. We should extract the `handleOnLoginWithGoogle` function and export it to use in both the login and signup screen.

lxkedinh (changes requested):

lxkedinh (approved):

PR #14, opened 2023-03-06
feat(prisma_migration): migrating from npm PostgreSQL client to Prisma

lxkedinh:
@GuyWhoCode and I migrated the repository's database client from [pg](https://www.npmjs.com/package/pg) to [Prisma](https://www.prisma.io/).

We migrated to Prisma because of its built-in intuitive database queries without needing us to write any SQL, and it abstracts away complex database relational models into simple Javascript objects for us to work with, improving the overall efficiency of our backend team to write new database-related code and debugging as well.


packages/server/controllers/auth.js
nappalion, line 1: Looks really good, and I love how Prisma removes SQL code. Does the 'authenticateWithGoogle' function need a try-catch?

lxkedinh: No try-catch is needed for that function since the API route handler that calls this function has the error handling.

nappalion (approved):

PR #17, opened 2023-03-16
feat(backend_testing_suite): Created Jest testing suite for our Express backend

lxkedinh:
I created a testing suite with [Jest](https://jestjs.io/) to encourage test-driven development with our backend team to write more robust and safe code.

Developers can create tests under `packages/server/__tests__/` and run tests with `yarn server:test` in the project root directory. There's an additional `setup-tests.js` file where the setup code for the backend testing environment should go such as starting the Express server and opening a database connection for testing queries.

To test our API endpoints, I added [Axios](https://axios-http.com/) for consistency since our frontend team also uses it.


packages/server/__tests__/setup-tests.js
```diff
@@ -0,0 +1,13 @@
...
+/**
+ * utility function to clean up and close database connection and express
+ * server after testing
+ */
+function cleanUpConnections() {
+  closePostgres();
+  server.close();
+}
+
+module.exports = cleanUpConnections;
```
GuyWhoCode, line 13: I am slightly confused on why this file is necessary. If there were multiple functions needed to initialize the tests to use Jest, it would be fine. Since only one function is created here, its indiviudal function calls can be made in `index.test.js` within the `afterAll`.

GuyWhoCode (approved):
Other than my previous comment, the rest of your code looks great!

packages/server/__tests__/setup-tests.js
lxkedinh: As you said, I wrote this utility file anticipating any new setup/teardown associated with possible dependencies we'll add to the project in the future. It'll be easier to do all of this in one function to be imported and used in Jest `beforeAll()` and `afterAll()` hooks. I only created `index.test.js` as a template for new team members to reference when creating new tests in the future like `events.test.js` and `users.test.js` for the events and users controllers.

dyl-tran (approved):
Looks good to me

PR #18, opened 2023-04-28
feat(coding_convention): added Prettier and ESLint for formatting and linting

lxkedinh:
[Prettier](https://prettier.io/) lets us format files automatically when saving (which should be set up by the developer) and before committing. [ESLint](https://eslint.org/) analyzes our codebase and looks for common coding patterns and errors for us.

Prettier needs to be properly integrated into your preferred editor to format files automatically. Developers can refer [here](https://prettier.io/docs/en/editors.html) to set that up.

I also added specific ESLint rules for React and React Native:
- [eslint-plugin-react](https://www.npmjs.com/package/eslint-plugin-react)
- [eslint-plugin-react-native](https://www.npmjs.com/package/eslint-plugin-react-native)

Regarding ESLint configuration, I added `.eslintrc.json` files in both our app and server repos, so we can have separate configurations for the front and backend when necessary.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,5 @@
+**/node_modules/
+.vscode/
+**/.expo/
+**/.expo-shared/
+**/assets/
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,3 @@
+{
+  "bracketSameLine": true
+}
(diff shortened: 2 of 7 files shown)


exelarios (approved):
LGTM

dyl-tran (approved):
Looks good to me, will use Prettier extension moving forward.

PR #21, opened 2023-06-09
feat(login): Implement refresh tokens

JustinNgxyen:
**Synopsis**:
Created refresh tokens that allow applicants to gain new access tokens without making the user re-login.

**Milestones**:
- The login route now creates both a refresh token and an access token when a user logins. The refresh token has a longer expiration date than the access token so that when the access token expires, the refresh token can create a new access token. 
- Logic of generating and verifying tokens moved to token.js
- The refresh and access tokens each have their own secret values from the environment variables.
- Each token will be verified according to their own separate secrets established in .env.
- The token route will make sure the refresh token is in the token list, as well as if the refresh token has been revoked, before updating the list with a new access token if the old one has expired.
- The revokeToken route will add the refresh token to a blacklist using Redis.

**Testing**:
- Postman was used to see if access and refresh tokens are created when a POST request to the login route is made using an email and password.
- The refresh token is used along with the email to create a POST request to the token route to test if an access token is created. 
- A GET request is made to the user route to test whether an access token contains the correct properties of the payload (user id, first name, last name, etc.). It also will see if the access token has expired.

RFC: https://docs.google.com/document/d/12DvNiTidBfp54obyKWnv8F2Bgs79_ZJzODNnsffAyzk/edit?usp=sharing


package.json
```diff
@@ -30,5 +30,9 @@
     "eslint-config-prettier": "8.8.0",
     "prettier": "2.8.8"
   },
-  "repository": "https://github.com/cppsea/icebreak.git"
+  "repository": "https://github.com/cppsea/icebreak.git",
+  "dependencies": {
+    "ioredis": "^5.3.2",
+    "redis": "^4.6.7"
+  }
```
lxkedinh: These dependencies shouldn't be here. Remember that this project is a monorepo that contains both the frontend and backend repos. Remove these and then install these dependencies in `packages/server`.

packages/server/routes/api/auth.js
```diff
@@ -50,19 +54,25 @@ router.post("/verify", async (request, response) => {
...
 });
 
 router.get("/user", AuthController.authenticate, (request, response) => {
   try {
+
+    const currentTimestamp = Math.floor(Date.now() / 1000);
+    if (request.user.exp < currentTimestamp) {
+      throw new Error("Access token has expired");
+    }
+
```
lxkedinh: This code to check expired tokens is unnecessary since the JWT package we're using already has expiration checking built-in. Try looking into incorporating that into the error handling for the `authenticate()` function in auth controllers.

https://github.com/auth0/node-jsonwebtoken#tokenexpirederror

packages/server/routes/api/auth.js
```diff
@@ -145,36 +158,56 @@ router.post("/register", async (request, response) => {
...
+        const postgres = new PostgresClient({
+          host: process.env.DB_HOST,
+          user: process.env.DB_USER,
+          database: process.env.DB_NAME,
+          password: process.env.DB_PASSWORD,
+          port: process.env.DB_PORT,
+        });
+
+        await postgres.connect(); 
+
```
lxkedinh: Why are you initializing postgres client here? The postgres client is initialized in a separate utility file and then imported everywhere we need it, so we don't reinitialize it every time we access the database.

packages/server/routes/api/auth.js
```diff
@@ -5,12 +5,16 @@ const token = require("../../utils/token");
 const { OAuth2Client } = require("google-auth-library");
 const bcrypt = require("bcrypt");
 const uniqid = require("uniqid");
+const tokenList = {};
```
lxkedinh: `tokenList` shouldn't be needed anymore in this file since we're using Redis to keep track of blacklisted tokens.

packages/server/routes/api/auth.js
```diff
@@ -196,24 +229,132 @@ router.post("/login", async (request, response) => {
...
+    }
+  } catch (error) {
+    response.status(500).send({
+      success: false,
+      message: error.message
+    });
+  }
+});
+
+router.post('/revokeToken', async (request, response) => {
```
lxkedinh: This should be `/revoke-token` to follow REST API naming convention. We should follow this when naming new API routes from now on, I'll add this to the resources channel.

image

https://restfulapi.net/resource-naming/

packages/server/routes/api/auth.js
```diff
@@ -196,24 +229,132 @@ router.post("/login", async (request, response) => {
...
+
+    const{ email, refreshToken } = request.body;
+    const requestedUser = await user.getUserByEmail(email);
+    // Check if refresh token is provided
+    if (refreshToken) {
+      // Verify the refresh token
+      token.verifyRefreshToken(refreshToken);
+      
+          // Check if the refresh token exists in tokenList
+          if (refreshToken in tokenList) {
```
lxkedinh: You should change this logic since we will be using Redis to reference against a blacklist of revoked tokens instead of checking if tokens are in `tokenList`.

packages/server/utils/token.js
```diff
@@ -1,16 +1,40 @@
...
-function verify(token) {
-  return jwt.verify(token, process.env.TOKEN_SECRET, { algorithms: ["HS256"] });
+function generateAccessToken(user){
+  return jwt.sign({
+    user_id: user.user_id,
+    joined_date: user.joined_date,
+    last_login: user.last_login,
+    first_name: user.first_name,
+    last_name: user.last_name,
+    avatar: user.avatar,
```
lxkedinh: We should try to keep the size of our tokens as small as possible. We should discuss with the frontend team to see which user properties are necessary to have in tokens and which ones are ok to leave out.

lxkedinh (changes requested):

packages/server/controllers/auth.js
```diff
@@ -102,9 +104,35 @@ async function deserialize(id, callback) {
...
+        response.status(401).json({ 
+          error: 'Invalid access token' 
+        });
+      }
+    } else {
+      // Token is valid
+      request.user = decoded;
+      next();
+    }
+  });
```
lxkedinh: Why do you have `token.verifyAccessToken(authToken);` and then verify the token again in the next line? You can move this logic to `verifyAccessToken` and throw errors where the `response.status().json()` methods are. That way, the errors you throw in `verifyAccessToken` will be caught and handled in this try/catch block instead.

lxkedinh (changes requested):

lxkedinh (approved):

PR #22, opened 2023-06-12
feat(group-screen): create group screen

PattyCakesxD:
Creates the UI for the Group Screen, including the header information and the contents of the sub-screens at the bottom of the page.

- Right now only supports the Events screen, which is basically a scrollview of EventCards. We're using test cards in this branch right now, but it should be easy to connect this to backend logic.
- There currently isn't a way to return to the Landing page via a back arrow icon. It's supposed to be on the top left of the page, but it's weird to implement when the header hides itself upon scrolling the EventsScreen, so I left out the implementation. Maybe this can be a future task?


lxkedinh:
I did another merge commit because, for some reason, your branch's files still weren't up to date after Patrick's last merge commit.

Overall, the screen looks really nice and the basic functionality is there, there are just some hiccups when the screen transitions from the banner when it's collapsing/expanding. Also, the current tab indicator bar under the tabs has a smooth animation sometimes or instantly snaps like in the video. Can you guys look into this?

https://github.com/cppsea/icebreak/assets/36950420/ab0c8593-2d12-46ab-9590-6255a6a81ecf

nappalion (approved):

PR #25, opened 2023-08-24
feat: Implementing Icebreak media asset storage with AWS S3

lxkedinh:
We implemented API routes and controllers to interact with our AWS S3 storage to store all our media assets (only images for now but potentially more types like videos later).

These API routes and controllers implement all the CRUD operations for our media assets


packages/server/controllers/images.js
```diff
@@ -0,0 +1,40 @@
...
+}
+
+async function retrieve(imageType, id) {
+  const key = imageType + "." + id + ".jpg";
+  const command = new GetObjectCommand({
+    Bucket: "icebreak-assets",
+    Key: key,
+  });
+
+  return await s3Client.send(command);
```
lxkedinh: For retrieving images, we want to get a signed URL to the image, not the actual JPEG image data itself which is what this `GetObjectCommand` is doing. You should return a signed URL like in your `upload` function.

packages/server/controllers/images.js
```diff
@@ -0,0 +1,40 @@
...
+    Body: body,
+  });
+  const getObjectCommand = new GetObjectCommand({
+    Bucket: "icebreak-assets",
+    Key: key,
+  });
+
+  await s3Client.send(putObjectCommand);
+  return await getSignedUrl(s3Client, getObjectCommand, {
+    expiresIn: 3600,
```
lxkedinh: This is a magic number, a number put directly in code that can be confusing because it's ambiguous and we might forget what it signifies later. It should be moved into a variable that should be used here instead to clarify what it is.

`const EXPIRATION_IN_SECONDS = 3600;` (preferably put this variable near the top of the file after the imports because it's a global static variable not meant to be modified)

and then replace with

`expiresIn: EXPIRATION_IN_SECONDS,`

packages/server/routes/api/images.js
```diff
@@ -0,0 +1,128 @@
...
+
+router.put("/:type", AuthController.authenticate, async (request, response) => {
+  const imageType = request.params.type;
+  const imageData = request.body.imageData;
+
+  if (
+    imageType !== "user_icon" &&
+    imageType !== "guild_icon" &&
+    imageType !== "guild_banner" &&
+    imageType !== "event_banner"
```
lxkedinh: These logical ANDs should be logical ORs, `||`. We want to return a fail response if the image type is not `user_icon` OR is not `guild_icon` OR etc.

packages/server/routes/api/images.js
```diff
@@ -0,0 +1,128 @@
...
+  ) {
+    response.status(400).json({
+      status: "fail",
+      message: "Image type '" + imageType + "' is invalid.",
+    });
```
lxkedinh: Instead of a message property, Fail responses should have a data object containing the property where the user error occurred. The property name should be corresponding to the parameter names. For this case, it should be something like

```javascript
{
  status: "fail",
  data: {
    imageType: `Image type ${imageType} is invalid.`,
  }
}
```

I won't request code reviews for all the other fail repsonses in your code, but you should change the others to follow this too. You can read the API specification docs we use [here](https://github.com/omniti-labs/jsend) for more clarification on the response format.

I also recommend looking into template literals which is what I did above with the string in `imageType`, one of the features of Javascript. It makes concatenating strings with variables easier than using `+`.

packages/server/routes/api/images.js
```diff
@@ -0,0 +1,128 @@
...
+  AuthController.authenticate,
+  async (request, response) => {
+    const imageType = request.params.type;
+    const id = request.params.id;
+
+    if (
+      imageType !== "user_icon" &&
+      imageType !== "guild_icon" &&
+      imageType !== "guild_banner" &&
+      imageType !== "event_banner"
```
lxkedinh: These should be logical ORs, `||`, instead of logical ANDs.

packages/server/routes/api/images.js
```diff
@@ -0,0 +1,128 @@
...
+        status: "fail",
+        message: "Image type '" + imageType + "' is invalid.",
+      });
+      return;
+    }
+
+    try {
+      const s3response = await ImagesController.retrieve(imageType, id);
+      response.status(200);
+      s3response.Body.pipe(response);
```
lxkedinh: I told you guys to do this method before we met up with Minh and finalized how we're going to implement retrieving images. You should change this to returning a signed url pointing to the image like you did in your uploading images route handler and `ImagesController.upload` function.

packages/server/routes/api/images.js
```diff
@@ -0,0 +1,128 @@
...
+    imageType !== "guild_banner" &&
+    imageType !== "event_banner"
+  ) {
+    response.status(400).json({
+      status: "fail",
+      message: "Image type '" + imageType + "' is invalid.",
+    });
+    return;
+  }
+  if (imageData === undefined) {
```
lxkedinh: A better condition would be
```javascript
if (!imageData) {
```

because your current condition doesn't cover the case of the empty string, "". In Javascript, `"" === undefined` is false.

packages/server/routes/api/images.js
```diff
@@ -0,0 +1,128 @@
...
+    });
+  } catch (err) {
+    response.status(500).json({
+      status: "error",
+      message: err.message,
+    });
+  }
+});
+
+router.get(
```
lxkedinh, line 95: You should also handle the case where an image with the supplied type and id were not found and return a fail response for that.

packages/server/routes/api/images.js
shotoh: No, think of it this way; if I had a guild icon, it would return true from imageType !== "user_icon", indicating a fail which we don't want

packages/server/routes/api/images.js
shotoh: My bad, I got mixed up with the error specification

packages/server/routes/api/images.js
HStevenC: I believe he meant if any fail, return fail status. The current condition requires all to fail to return fail status. By fail I meant true from imageType !== "type."

HStevenC: Since it's a bit convoluted. We can have 
`const validImageTypes = [
  "user_icon",
  "guild_icon",
  "guild_banner",
  "event_banner"
];`
at the top and use `if (!validImageTypes.includes(imageType)) `

packages/server/controllers/images.js
```diff
@@ -0,0 +1,40 @@
+const { v4: uuidv4 } = require("uuid");
+const { getSignedUrl } = require("@aws-sdk/s3-request-presigner");
+const { s3Client } = require("../utils/s3");
+const { PutObjectCommand, GetObjectCommand } = require("@aws-sdk/client-s3");
+
+async function upload(imageType, imageData) {
+  const id = uuidv4();
```
lxkedinh: You should not generate a new UUID here. The UUID should come from the frontend in the request body along with the image type and data. Pass the UUID as a parameter to the upload function here to use for the key.

packages/server/routes/api/images.js
shotoh: Alright I'll implement that

packages/server/routes/api/images.js
```diff
@@ -0,0 +1,243 @@
...
+
+const AuthController = require("../../controllers/auth");
+const ImagesController = require("../../controllers/images");
+
+const validImageTypes = [
+  "user_icon",
+  "guild_icon",
+  "guild_banner",
+  "event_banner",
+];
```
lxkedinh: There's no convention for these types of global constant variables, but it would be a good idea to use all uppercase with underscores to differentiate them from other regular variables in our code. Something like
```javascript
const VALID_IMAGE_TYPES = [
  "user_icon",
  "guild_icon",
  "guild_banner",
  "event_banner",
];
```

packages/server/routes/api/images.js
```diff
@@ -0,0 +1,243 @@
...
+    }
+    // if (!validate(imageUUID)) {
+    //   response.status(400).json({
+    //     status: "fail",
+    //     data: {
+    //       imageUUID: `Image UUID '${imageUUID}' is invalid.`,
+    //     },
+    //   });
+    //   return;
+    // }
```
lxkedinh: Uncomment this code and make sure the invalid provided UUID exception is handled.

packages/server/routes/api/images.js
```diff
@@ -0,0 +1,243 @@
...
+  }
+);
+
+router.delete( "/:type/:id", 
+//AuthController.authenticate,
+ async (request, response) => {
+    const imageType = request.params.type;
+    const id = request.params.id;
+    if (!validImageTypes.includes(imageType)) {
+      response.status(403).json({
```
lxkedinh: 403 is not the correct HTTP status code to use here. 400 would be a better option. See [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400). This goes for the other fail responses due to invalid user input as well.

packages/server/routes/api/images.js
```diff
@@ -0,0 +1,243 @@
...
+    //     status: "fail",
+    //     data: {
+    //       imageUUID: `Image UUID '${imageUUID}' is invalid.`,
+    //     },
+    //   });
+    //   return;
+    // }
+    try {
+      const deleteResponse  = await ImagesController.Delete(imageType, id);
+      console.log(deleteResponse);
```
lxkedinh: Having `console.log()` is fine for now but don't forget to remove these when we finish and merge this to main branch.

packages/server/routes/api/images.js
```diff
@@ -0,0 +1,243 @@
...
+      status: "fail",
+      message: `Image type ${imageType} is invalid.`,
+    });
+    return;
+  }
+  if (!imageData) {
+    response.status(400).json({
+      status: "fail",
+      message: "Request body missing values 'id' and/or 'imageData'.",
+    });
```
lxkedinh: Like I told Andrew, remember that the format of fail type responses has a data property with a key of the parameter that went wrong. See [here](https://github.com/omniti-labs/jsend#fail) for more explanation.

packages/server/controllers/images.js
```diff
@@ -0,0 +1,73 @@
...
+    Bucket: "icebreak-assets",
+    Key: key,
+  });
+
+  return await getSignedUrl(s3Client, command, {
+    expiresIn: EXPIRATION_IN_SECONDS,
+  });
+}
+
+async function Delete(imageType, id) {
```
lxkedinh: The rest of the codebase follows camelcase naming convention so change this to `delete` to follow that.

packages/server/controllers/images.js
```diff
@@ -0,0 +1,73 @@
...
+  const key = id + ".jpg";
+  //const key = imageType + "." + id + ".jpg"; UUID
```
lxkedinh: The current key used is incorrect. The commented out lines for the keys in both your delete and patch function are correct and should be used instead.

packages/server/routes/api/images.js
```diff
@@ -0,0 +1,243 @@
...
+    } catch (err) {
+      response.status(500).json({
+        status: "error",
+        message: err.message,
+      });
+    }
+  }
+);
+
+router.delete( "/:type/:id",
```
lxkedinh: I'm seeing different route parameters between the route handlers, `:id` or `:UUID`. The RFC has `:UUID` so the route parameter should be changed to have `:UUID` instead.

packages/server/controllers/images.js
```diff
@@ -0,0 +1,73 @@
...
+  const command = new DeleteObjectCommand({
+    Bucket: "icebreak-assets",
+    Key: key,
+  });
+  const response = await s3Client.send(command);
+  return response;
+}
+
+async function patch(imageType, imageData, id) {
```
lxkedinh: When naming your functions, you want to make them both concise, clear, and explicit. We don't have to name the controller functions after the HTTP methods we are using as well (PUT, PATCH, DELETE). When someone sees `ImageController.patch()`, it might be a little unclear. What does patch mean exactly? `ImageController.update()` is a little more explicit, so changing this function to `update()` would be a good idea.

packages/server/routes/api/images.js
```diff
@@ -0,0 +1,243 @@
...
+  }
+  // if (!validate(imageUUID)) {
+    //   response.status(400).json({
+    //     status: "fail",
+    //     data: {
+    //       imageUUID: `Image UUID '${imageUUID}' is invalid.`,
+    //     },
+    //   });
+    //   return;
+    // }
```
lxkedinh: Uncomment this as it covers the invalid UUID exception case.

packages/server/controllers/images.js
HStevenC: lowercase is a reserved word

package-lock.json
lxkedinh, line 1: Remember that we only use **yarn** when syncing/updating dependencies, **never use npm**. This package-lock.json file is created when you add dependencies with npm and it can break our existing dependencies.

packages/server/routes/api/images.js
```diff
@@ -0,0 +1,196 @@
...
+    }
+    if (!imageData) {
+      response.status(400).json({
+        status: "fail",
+        data: {
+          imageData: "Request body missing value 'imageData'.",
+        },
+      });
+      return;
+    }
```
lxkedinh, line 49: I didn't put this in the RFC so that's my bad, but we should also check to make sure the Base64 string we receive is actually encoded JPEG data. This goes for the imageData validation checks in the other route handlers, too. I added details in the RFC on how to implement this.

lxkedinh (commented):
I updated this branch with changes from main branch for you guys, so you guys should figure out how to use Prisma to interact with our database and actually query for the database entity we're trying to upload the image for.

packages/server/routes/api/images.js
```diff
@@ -151,6 +151,13 @@ router.patch(
     const imageType = request.params.type;
     const imageUUID = request.params.UUID;
     const imageData = request.body.imageData;
+    if (!imageData.startsWith("/9j/")){
+      response.status(400).json({
+        status: "fail",
+        message: `Invalid Base64-encoded JPEG`,
+      });
+      return;
+    }
```
lxkedinh: Don't forget about the format of fail-type API responses. It should be
```javascript
response.status(400).json({
  status: "fail",
  data: {
    imageData: "Invalid Base64-encoded JPEG"
  }
});
```

We also need this check on the image upload API route handler.

packages/server/routes/api/images.js
lxkedinh, line 1: We'll need to check for entity existence in the database for the updating images PATCH route handler as well since updating an image is the same as uploading one in AWS S3.

packages/server/routes/api/images.js
```diff
@@ -0,0 +1,232 @@
...
+    const imageType = request.params.type;
+    const imageUUID = request.params.UUID;
+    const imageData = request.body.imageData;
+    if (!imageData.startsWith("/9j/")) {
+      response.status(400).json({
+        status: "fail",
+        message: `Invalid Base64-encoded JPEG`,
+      });
+      return;
+    }
```
lxkedinh: This check will have to go after the `imageData` null check otherwise this can crash the API if the frontend doesn't put anything for `imageData`.

packages/server/routes/api/images.js
```diff
@@ -0,0 +1,243 @@
...
+          imageUUID: `Image UUID '${imageUUID}' is invalid.`,
+        },
+      });
+      return;
+    }
+
+    try {
+      const s3response = await ImagesController.retrieve(imageType, imageUUID);
+      response.status(200);
+      response.send(s3response);
```
lxkedinh: Remember to follow the API response format. This is a success API response so the response sent back to the frontend should be
```javascript
{
  status: "success",
  data: {
    imageUrl: s3Response
  }
}
```

packages/server/routes/api/images.js
```diff
@@ -0,0 +1,243 @@
...
+    }
+    if (!(await ImagesController.exists(imageData, imageUUID))){
+      response.status(400).json({
+        status: "fail",
+        data: {
+          imageUUID: `A ${imageType} with UUID ${imageUUID} does not exist.`,
+        },
+      });
+      return;
+    }
```
lxkedinh: you are passing the wrong variable here in `ImagesController.exists` in the if statement condition. It should be `ImagesController.exists(imageType, imageUUID)`.

packages/server/routes/api/images.js
lxkedinh, line 1: Reminder to get rid of any console.logs you guys put in the code for testing while working on this before we squash and merge to main.

packages/server/routes/api/images.js
lxkedinh: I tested fetching a nonexistent image in the S3 storage and it gave me a dead link and was still considered a successful response. The API route should throw an error if the frontend tries to fetch a nonexistent image with the GET route. You'll have to change `ImagesController.retrieve` to throw an error if there is no image in S3 that exists with the given key.

HStevenC (approved):

PR #26, opened 2023-08-31
Typescript Migration - Backend

mikephu:
- Currently working on migrating controllers and api routing under packages/server to typescript.


packages/server/controllers/guilds.ts
```diff
@@ -0,0 +1,21 @@
+import prisma from "../utils/prisma";
+import { Guild } from ".prisma/client/index";
+
+async function getAllGuilds() {
+  const query = await prisma.guild.findMany();
+  return query;
+}
```
lxkedinh: We should write down the return type annotations for all the controller functions for clarity.

packages/server/controllers/events.ts
```diff
@@ -1,16 +1,17 @@
 // const postgres = require("../utils/postgres");
-const prisma = require("../utils/prisma");
+import prisma from "../utils/prisma";
+import { Event } from ".prisma/client/index";
```
lxkedinh: your imports for the `Event` type and `Guild` type in the `guilds.ts` file seem to be broken. They showed up as errors on my editor and the correct import was
```javascript
import { Event } from "@prisma/client";
import { Guild } from "@prisma/client";
```

These types are automatically generated by the Prisma client so make sure you have Prisma client generated by running `yarn workspace server prisma generate` in icebreak root directory.

lxkedinh (changes requested):
Really good job so far Michael, there are only a few minor things that need to be changed.

PR #30, opened 2023-11-13
(feat) Guild Members API

JustinNgxyen:
**Objective**: Each guild has a collection of members, and the primary goal of this API route is to retrieve information about all users associated with a specific guild.

- Data Retrieval of all users (usernames, email addresses, avatar URLs, etc.)
- Authentication: Only authenticated and authorized users or applications with valid tokens should be able to access the guild user data.
- Error Handling

**Implementation**: A GET route in guild.js that uses the guildID to retrieve the guild users using the GuildController function getGuildUsers.
If the guildId is undefined, either not inputted or invalud, than a fail status is sent along with the reasoning.


packages/server/routes/api/guilds.js
```diff
@@ -53,4 +53,35 @@ router.get(
...
+        });
+      }
+
+      const users = await GuildController.getGuildUsers(guildId);
+      response.status(200).json({
+        status: "success",
+        data: {
+          users,
+        },
+      });
```
lxkedinh: Changing `users` to something more explicit like `guildMembers` would be better and add more clarity and readability.

packages/server/controllers/guilds.js
```diff
@@ -14,7 +14,17 @@ async function getGuild(guildId) {
   return query;
 }
 
+async function getGuildUsers(guildId) {
```
lxkedinh: Change this function name to be more explicit like `getGuildMembers`.

packages/server/controllers/guilds.js
```diff
@@ -14,7 +14,17 @@ async function getGuild(guildId) {
...
+  const query = await prisma.guildMembers.findMany({
+    where: {
+      guildId: guildId,
+    },
+  });
+  return query;
```
lxkedinh: Right now, this query returns the user UUID and guild UUID for all the members of a guild. Since frontend wants to display members in a list, we want the Prisma query to fetch
- user's UUID
- user's first name
- user's last name

Your original idea for the Prisma query was more correct than what I said to change it to so my bad on that. After looking into it more [here](https://www.prisma.io/docs/concepts/components/prisma-schema/relations/many-to-many-relations#explicit-many-to-many-relations), we want to query the users table passing the guild UUID but it's different from a regular query. It will look something like this. 

image

lxkedinh (changes requested):
looks good to me so far, just needs some minor changes

packages/server/controllers/guilds.js
```diff
@@ -14,7 +14,29 @@ async function getGuild(guildId) {
...
+    },
+    select: {
+      members: {
+        select: {
+          userId: true,
+          firstName: true,
+          lastName: true,
+        },
+      },
+    },
```
lxkedinh: We also need to get the avatar icon for each guild member to display in the list.

packages/server/routes/api/guilds.js
lxkedinh, line 1: Two checks are missing from your route handler.
- Check to make sure `guildId` is an actual UUID.
- Check to make sure `guildId` belongs to an actual guild that exists in the database.

lxkedinh (changes requested):

lxkedinh (approved):

PR #31, opened 2023-11-21
Guild crud api routes

samlyme:
Implementing CRUD operations for guilds.


packages/server/routes/api/guilds.js
```diff
@@ -4,51 +4,166 @@ const router = express.Router();
...
+            errorName: error.name,
+            errorMessage: "The guildId Provided had invalid format.",
+          });
+          
+        case "NotFoundError":
+          return response.status(500).json({
+            status: "fail",
+            errorName: error.name,
+            errorMessage: error.message,
+          });
```
lxkedinh: This error is deprecated in Prisma so we shouldn't rely on it. We should rely on checking for `error.code === 'P2025'` for database records not found instead. I gave an example in a code snippet in my other comment about narrowing the error handling even more.
image

packages/server/routes/api/guilds.js
```diff
@@ -4,51 +4,166 @@ const router = express.Router();
...
+      
+    } catch (error) {
+      switch (error.name) {
+
+        case "PrismaClientKnownRequestError":
+          return response.status(400).json({
+            status: "error",
+            errorName: error.name,
+            errorMessage: "The guildId Provided had invalid format.",
+          });
```
lxkedinh: This is a good start on narrowing the errors that are caught. However, `PrismaClientKnownRequestError` itself is still a broad umbrella error type that encompasses more specific errors. In addition to checking for `PrismaClientKnownRequestError`, we can check `error.code` to narrow it even more. [Prisma docs](https://www.prisma.io/docs/reference/api-reference/error-reference#error-codes) have a detailed list of all their error codes. Changing your error handling to something like this would be good.

```js
if (error.name === "PrismaClientKnownRequestError") {
  switch (error.code) {
    case "P2023":
      return response.status(400).json({
        status: "fail",
        data: {
          guildId: "Provided guild ID is not a UUID."
        }
      });
      break;
    case "P2025":
      return response.status(404).json({
        status: "fail",
        data: {
          guildId: `No guild with an ID of ${guildId} could be found.`
        }
      });
      break;
      // any other error codes we want to handle and send back specific error messages for
  }
}
```

I won't comment on the error handling in the other routes, but adjust those to add this change as well.

packages/server/routes/api/guilds.js
```diff
@@ -4,51 +4,166 @@ const router = express.Router();
...
+    // Should never error.
+    // In the case that there are no guilds, 
+    // the return would be an empty array.
+    // Here for redudancy.
+    response.status(400).json({
       status: "error",
-      message: error.message,
+      errorName: error.name,
+      errorMessage: error.message,
     });
```
lxkedinh: Make sure to follow the [JSend](https://app.clickup.com/9011073139/v/dc/8chm43k-2311/8chm43k-2871) format for your API responses. I won't comment on the other ones in your code but make sure they follow the format for the sake of consistency across our entire API.

For this response, it should be changed to
```js
response.status(500).json({
  status: "error",
  message: error.message
});
```

Also for HTTP status code numbers, the 400 range is usually reserved for a "bad request" when the client making the API request makes an invalid request like invalid data. The 500 range is usually reserved for errors that occur server-side.

packages/server/routes/api/guilds.js
```diff
@@ -4,51 +4,166 @@ const router = express.Router();
...
+          });
       }
+    }
+  }
+);
 
-      const guild = await GuildController.getGuild(guildId);
-      response.status(200).json({
+// Create Guild
+router.post(
```
lxkedinh: Right now, you have `PUT` for updating guilds and `POST` for creating guilds. It should be the other way around, `PUT` is associated with creating new resources in REST API convention, and `POST` is associated with updating resources.

packages/server/__tests__/controllers/guilds.test.js
```diff
@@ -0,0 +1,201 @@
...
+  };
+
+  const initialMultipleUpdateGuildResponseOneField = {
+    name: "Facebook",
+  };
+
+  const expectedUpdateGuildResponseOneField = {
+    guildId: "308b5bc4-7771-4514-99b7-9611cefb7e1d",
+    name: "Facebook",
+  };
```
lxkedinh: These variable names are starting to get really long. Since some of these are only used for one specific test, you can declare them inside each test block it is used in instead and keep the variable names shorter like `expected`.

For example for your "should update multiple guild properties successfully" test, something like this:
```js
test("should update multiple guild properties successfully", async () => {
  // this replaces expectedUpdateGuildResponse
  const expected = {
    guildId: "308b5bc4-7771-4514-99b7-9611cefb7e1d",
    name: "Apple",
    handler: "Tim Cook",
    description: "Come buy our iphones.",
    category: "Guild Category",
    location: "Optional Location",
    website: "Optional Website",
    tags: ["Tag1", "Tag2", "apple 15"],
    banner: "URL to Banner Image",
    icon: "URL to Icon Image",
    media: ["Media1", "Media2", "Media3"],
    isInviteOnly: true,
  };

  prismaMock.guilds.update.mockResolvedValue(expected);

  const result = await GuildsController.updateGuild(
    "308b5bc4-7771-4514-99b7-6666cefb7e1d",
    initialGuild // originally initialMultipleUpdateGuildResponse, replaced name because it was too long
  );

  expect(result).toEqual(
    expect.objectContaining(expectedUpdateGuildResponse)
  );
});
```

packages/server/__tests__/controllers/guilds.test.js
```diff
@@ -0,0 +1,201 @@
...
+
+  const guildIdToDelete = "308b5bc4-7771-4514-99b7-6666cefb7e1d";
+
+  test("should successfully delete a target guild", async () => {
+    prismaMock.guilds.delete.mockResolvedValue(guildIdToDelete);
+
+    const result = await GuildsController.deleteGuild(guildIdToDelete);
+
+    expect(result).toEqual(guildIdToDelete);
+  });
```
lxkedinh: According to Prisma docs, when deleting from the database, the `delete()` query returns the object that was deleted. Change your mock resolved value to a mock guild object instead.

image

lxkedinh (changes requested):
You guys are off to a good start! Most of the things I'm requesting changes for are just convention stuff, but overall the basic functionality is there.

packages/server/routes/api/guilds.js
larrylaa: Please double check this comment.

packages/server/__tests__/controllers/guilds.test.js
```diff
@@ -0,0 +1,175 @@
...
+} = require("@prisma/client/runtime/library");
+const GuildsController = require("../../controllers/guilds");
+const { prismaMock } = require("../prisma_mock");
+
+describe("Guilds Unit Tests", () => {
+  /*
+  Unit Tests: Fetch Guild By ID
+  - Successfully fetching guild by ID ✅
+  - Fetching a guild with an invalid ID that doesn't exist in database (should throw error) ✅
+  */
```
lxkedinh: Ideally, unit tests should be self-documenting. If you have good test names, people not familiar with your code should be able to completely understand what your tests accomplish. Having these comments is redundant.

packages/server/__tests__/controllers/guilds.test.js
```diff
@@ -0,0 +1,175 @@
...
+
+  test("should find no guild if given nonexistent id", async () => {
+    prismaMock.guilds.findUniqueOrThrow.mockRejectedValue(
+      new PrismaClientKnownRequestError("record not found", { code: "P2025" })
+    );
+
+    await expect(
+      GuildsController.getGuild("e2bdade9-4bf2-4220-8020-e09266363762")
+    ).rejects.toThrow(PrismaClientKnownRequestError);
+  });
```
lxkedinh: For the tests you write that intentionally fail and you expect them to throw, you should include that in the test name for clarity, like `should throw if no guild has given id`.

packages/server/__tests__/controllers/guilds.test.js
```diff
@@ -0,0 +1,175 @@
...
+      website: "apple.com",
+      banner: "apple.jpeg",
+      icon: "appleicon.jpeg",
+    };
+
+    prismaMock.guilds.create.mockResolvedValue(
+      new PrismaClientKnownRequestError("Missing required field error", {
+        code: "P2012",
+      })
+    );
```
lxkedinh: You should use `mockRejectedValue(...)` instead of `mockResolvedValue(...)` for throwing errors in tests that intentionally fail.

packages/server/__tests__/controllers/guilds.test.js
```diff
@@ -0,0 +1,175 @@
...
+
+    await expect(
+      GuildsController.createGuild(testinvalidCreateGuild)
+    ).resolves.toThrow(PrismaClientKnownRequestError);
+  });
```
lxkedinh: This should be `expect(...).rejects.toThrow(...)`. Use `.resolves` when you expect the controller function to "succeed" in the test and use `.rejects` when you expect the controller function to "fail" in the test and throw. `.toThrow(...)` is only used with `.rejects`.

packages/server/routes/api/guilds.js
```diff
@@ -26,29 +27,179 @@ router.get(
...
+      // Note: Keep this here until Prisma NotFoundError is fully deprecated with the future version of the client, as of now, it will not reach the P2025 case in the conditional below.
+      if (error.code === "P2025") {
+        return response.status(404).json({
           status: "fail",
           data: {
-            guildId: "Guild ID not provided",
+            guildId: `No guild with an ID of ${request.params.guildId} could be found.`,
           },
         });
       }
```
lxkedinh: This if statement is unnecessary because the switch statement right below this line is already covering `error.code = "P2025"`. Only check for `error.code` for Prisma Client errors otherwise it could be undefined.

packages/server/routes/api/guilds.js
```diff
@@ -26,29 +27,179 @@ router.get(
...
+      },
+    });
+  } catch (error) {
+    // Note: When errored, this route returns a huge error message, the end of the message contains missing arugments/fields required for the route to function properly.
+    return response.status(400).json({
+      status: "error",
+      errorName: error.name,
+      message: error.message,
+    });
+  }
```
lxkedinh: This should be status code 500 for default error handling.

packages/server/routes/api/guilds.js
```diff
@@ -26,29 +27,179 @@ router.get(
...
+router.post("/", AuthController.authenticate, async (request, response) => {
+  try {
+    return response.status(200).json({
+      status: "success",
+      data: {
+        createdGuild: await GuildController.createGuild(request.body),
+      },
+    });
+  } catch (error) {
+    // Note: When errored, this route returns a huge error message, the end of the message contains missing arugments/fields required for the route to function properly.
```
lxkedinh: When a client tries to create a guild with invalid data, the entire Prisma error message is sent back as a response. Add more error handling to check for `error.name` of `PrismaClientValidationError` and then send an error message back like "Guild could not be created because of missing or invalid arguments`

packages/server/routes/api/guilds.js
```diff
@@ -26,29 +27,179 @@ router.get(
...
+
+          default:
+            return response.status(500).json({
+              status: "fail",
+              errorName: error.name,
+              errorMessage: error.message,
+            });
+        }
+      } else {
+        return response.status(500).json({
```
lxkedinh: Guild updating also needs similar error handling like I said for the guild creation route. Check for `error.name` of `PrismaClientValidationError` and then send an error message back like "Guild could not be updated because of missing or invalid arguments`.

packages/server/routes/api/guilds.js
```diff
@@ -4,19 +4,20 @@ const router = express.Router();
...
       },
     });
   } catch (error) {
-    response.status(500).json({
+    response.status(400).json({
       status: "error",
-      message: error.message,
+      errorName: error.name,
+      errorMessage: error.message,
     });
```
lxkedinh: For error API responses, no need for an `errorName` property. Stick to this format. This goes for all the error API responses you have for the guild routes.
```js
{
  status: "error",
  message: error.message
}
```

packages/server/routes/api/guilds.js
```diff
@@ -26,29 +27,179 @@ router.get(
...
+                guildId: `No guild with an ID of ${request.params.guildId} could be found.`,
+              },
+            });
+
+          default:
+            return response.status(500).json({
+              status: "fail",
+              errorName: error.name,
+              errorMessage: error.message,
+            });
```
lxkedinh, line 73: This default case is unnecessary since it's just a copy of the default error handler and this goes for the other routes with switch statements in error handling too. Keep this if statement but remove the else block wrapping the default error response below this line. Since your switch cases have return statements, the default error handler below won't run if the switch cases are executed.

lxkedinh (changes requested):

packages/server/routes/api/guilds.js
```diff
@@ -47,31 +48,163 @@ router.get("/", AuthController.authenticate, async (request, response) => {
...
+  } catch (error) {
+    if (error.name === "PrismaClientValidationError") {
+      return response.status(400).json({
+        status: "fail",
+        data: {
+          guildId:
+            "Guild could not be created because of missing or invalid arguments, see last line of error message below for details.",
+          message: error.message,
+        },
+      });
```
lxkedinh: We're only keeping the `guildId` message, we don't want to send the entire Prisma error message since sometimes they can have sensitive info. Having just a `"Guild could not be created because of missing or invalid arguments"` is enough for a clear but generic message on what went wrong. This goes for the other `PrismaClientValidationError` check in the updating guilds route as well.

packages/server/routes/api/guilds.js
```diff
@@ -47,31 +48,163 @@ router.get("/", AuthController.authenticate, async (request, response) => {
...
+      const guild = await GuildController.getGuild(guildId);
+
+      return response.status(200).json({
+        status: "success",
+        data: {
+          guild,
+        },
+      });
+    } catch (error) {
+      if (error.name === "PrismaClientKnownRequestError") {
```
lxkedinh: I did some more looking into [Prisma docs](https://www.prisma.io/docs/orm/prisma-client/debugging-and-troubleshooting/handling-exceptions-and-errors) and they recommend using `error instanceof PrismaClientKnownRequestError` to check if the error is an instance of one of the Prisma error subclasses instead of checking `error.name` but you guys got the general idea down with error handling.

lxkedinh (commented):
Looks good to me guys! There were only a few things over semantics that I took care of myself but functionality-wise everything looks great. Really good job!

lxkedinh (approved):

PR #32, opened 2023-12-03
feat: get event attendees

HStevenC:


packages/server/routes/api/events.js
```diff
@@ -114,4 +114,48 @@ router.get(
...
+        },
+      });
+      return;
+    }
+    if (!(await EventController.existsInPrisma(eventId))) {
+      response.status(404).json({
+        status: "fail",
+        data: {
+          eventId: `Given eventId "${eventId}" does not exist.`,
+        },
```
lxkedinh: The error message here would be clearer with something like `No event exists with an ID of ${eventId}`.

packages/server/routes/api/events.js
```diff
@@ -114,4 +114,48 @@ router.get(
   }
 );
 
+router.get(
+  "/eventAttendees/:eventId",
```
lxkedinh: REST API naming says to use a hierarchy for the resources we provide in the API. This route will be on `/events`, and thinking from a design perspective, the attendees belong to the event with `:eventId`, so the route path here should be something like `/:eventId/attendees`.

lxkedinh (changes requested):
Code looks really good so far! Only minor changes needed and this PR can be finished up soon

lxkedinh (approved):

PR #34, opened 2023-12-19
Event crud routes

BLiu456:
Implementation of API routes for events.


packages/server/controllers/events.js
```diff
@@ -52,6 +52,23 @@ async function getPages(limit) {
...
+      description: eventData.description,
+      startDate: eventData.startDate,
+      endDate: eventData.endDate,
+      location: eventData.location,
+      thumbnail: eventData.thumbnail,
+    },
+  });
+  return newEvent;
+}
+
```
lxkedinh: When preparing to open a PR ready for review, make sure to go through your code again and clean it up. Get rid of any commented-out code, console logs you did for debugging, etc.

packages/server/controllers/events.js
```diff
@@ -61,9 +78,47 @@ async function getEvent(eventId) {
...
+    const deletedEvent = await prisma.events.delete({
+      where: {
+        eventId: eventId,
+      },
+    });
+    return deletedEvent;
+  } catch (error) {
+    return null;
+  }
+}
```
lxkedinh: You don't need a try-catch here because the delete event route handler already has a try-catch. Just return `deletedEvent` and let the try-catch inside the delete event route handler do the error handling.

packages/server/controllers/events.js
```diff
@@ -61,9 +78,47 @@ async function getEvent(eventId) {
...
+        startDate: eventData.startDate,
+        endDate: eventData.endDate,
+        location: eventData.location,
+        thumbnail: eventData.thumbnail,
+      },
+    });
+    return updateEvent;
+  } catch (error) {
+    return null;
+  }
```
lxkedinh: Same thing as earlier comment, the try-catch here is unnecessary because the try-catch inside its corresponding route handler will do the error handling.

packages/server/routes/api/events.js
```diff
@@ -92,6 +92,120 @@ router.get(
...
+    if (givenData.title) {
+      if (givenData.title.length > 255) {
+        response.status(400).json({
+          status: "fail",
+          data: {
+            title: "Title exceeds 255 characters",
+          },
+        });
+      }
+    }
```
lxkedinh: I'm not going to highlight the whole thing, but this comment applies to all the if statements below this one as well. When returning an API response with `response.status().json()`, you still need to return from the route handler function, otherwise the route handler will keep going through the function. Each of these if statements will need a return to stop execution if user input is invalid.

packages/server/routes/api/events.js
lxkedinh, line 1: I'm looking into possibly adding [express-validator](https://express-validator.github.io/docs), a library for input validation and sanitization. These if-statement guard clause checks are starting to clutter up the route handlers.

lxkedinh, line 1: You guys are missing some important input validation checks for the following route handlers.

Creating events:
- guildId is null (return fail API response with message that input guildId cannot be null)
- guildId is not a UUID (return fail API response with message that input guildId is not a UUID)
- guildId is nonexistent, no guilds in the database have the guildId (return fail API response with message that no guild exists with input guildId)
- title is null (return fail API response with message that titles are required for creating events)

Updating events:
- eventId is null (return fail API response with message that input eventId is required for updating events)
- eventId is not a UUID (return fail API response with message that eventId is not a UUID)
- eventId is nonexistent, no events in the database have that eventId (return fail API response with message that no event exists with input eventId)

Deleting events:
- eventId is null (return fail API response with message that input eventId is required for updating events)
- eventId is not a UUID (return fail API response with message that eventId is not a UUID)
- eventId is nonexistent, no events in the database have that eventId (return fail API response with message that no event exists with input eventId)

All of the above should have status code 400 for Bad Request because the client messed up in their request somehow.

packages/server/routes/api/events.js
```diff
@@ -114,4 +228,179 @@ router.get(
...
+        message: error.message,
+      });
+    }
+  }
+);
+
+router.post(
+  "/:eventId",
+  AuthController.authenticate,
+  async (request, response) => {
```
lxkedinh: Swap the POST and PUT. I was mistaken about the standard convention for this.
- POST is for creating
- PUT is for updating

lxkedinh (changes requested):
There are a few things that need to be addressed, but this is a really good job for your guys' first task.

packages/server/validators/events.js
```diff
@@ -0,0 +1,229 @@
...
+    .exists({ checkFalsy: true })
+    .withMessage("guildId cannot be null or empty")
+    .isUUID()
+    .withMessage("Not a UUID")
+    .custom(async (value) => {
+      try {
+        await GuildController.getGuild(value); //Check if guild exists by getting it
+      } catch (error) {
+        //If event does not exists, then a NotFoundError is thrown from controller
+        throw new Error(error);
```
lxkedinh: Instead of throwing the Prisma error, throw a new error with a generic but descriptive message of what went wrong. We don't want to throw the Prisma error because it can contain sensitive info. Something like
```js
throw new Error(`No guild exists with an ID of ${value}`
```
would be good. This goes for the other custom validators that involve Prisma queries as well.

packages/server/validators/events.js
```diff
@@ -0,0 +1,229 @@
...
+    .withMessage("Title length must be between 1 to 255 characters"),
+
+  //Description checks
+  body("description", "Invalid description")
+    .trim()
+    .escape()
+    .isLength({ max: 255 })
+    .withMessage("Description max length is 255 characters")
+    .exists({ checkFalsy: true })
+    .withMessage("Description cannot be null or empty"),
```
lxkedinh: Double-check the event property checks on your validators, the only non-null event properties are eventId, guildId, and title. You don't need non-null checks on the other ones.

packages/server/validators/events.js
```diff
@@ -0,0 +1,229 @@
...
+    .withMessage("Title cannot be null or empty")
+    .isLength({ min: 1, max: 255 })
+    .withMessage("Title length must be between 1 to 255 characters"),
+
+  //Description checks
+  body("description", "Invalid description")
+    .trim()
+    .escape()
+    .isLength({ max: 255 })
+    .withMessage("Description max length is 255 characters")
```
lxkedinh: 255 is way too small of a character limit for descriptions. Think of something along the lines of LinkedIn, Instagram, and Facebook posts. It should be around 2,000 maybe 3,000 characters.

packages/server/validators/events.js
```diff
@@ -0,0 +1,229 @@
...
+    .exists({ checkFalsy: true })
+    .withMessage("Event Id cannot be null or empty")
+    .custom(async (value) => {
+      try {
+        await EventController.getEvent(value); //Check if event exists by getting it
+      } catch (error) {
+        //If event does not exists, then a NotFoundError is thrown from controller
+        throw new Error(error);
+      }
+    }),
```
lxkedinh: This is a duplicate check from the `eventIdValidator` already defined at the bottom of the page. You can get rid of this and instead use both validators in the update event route handler. It would look something like this
```js
router.put(
  "/:eventId",
  AuthController.authenticate,
  eventIdValidator,
  updateEventValidator,
  //  ...
});
```

packages/server/routes/api/events.js
```diff
@@ -92,6 +98,46 @@ router.get(
...
+        status: "error",
+        message: error.message,
+      });
+    }
+  }
+);
+
 router.get(
   "/:eventId",
   AuthController.authenticate,
```
lxkedinh: You guys are missing a validator for the getting an event by ID route.

packages/server/routes/api/events.js
```diff
@@ -114,4 +160,91 @@ router.get(
...
+      const { eventId } = request.params;
+      //See what data was given in request. Any data not given to be updated will be undefined
+      let givenData = {
+        title: request.body.title || undefined,
+        description: request.body.description || undefined,
+        startDate: request.body.startDate || undefined,
+        endDate: request.body.endDate || undefined,
+        location: request.body.location || undefined,
+        thumbnail: request.body.thumbnail || undefined,
+      };
```
lxkedinh: Update this to use the validated data from your validator. See [here](https://express-validator.github.io/docs/guides/getting-started#accessing-validated-data).

lxkedinh (changes requested):

packages/server/validators/events.js
BLiu456: Updating the description now goes up to a max of 2000 characters.

packages/server/validators/events.js
```diff
@@ -0,0 +1,203 @@
...
+    .withMessage("Title cannot be null or empty")
+    .isLength({ min: 1, max: 255 })
+    .withMessage("Title length must be between 1 to 255 characters"),
+
+  //Description checks
+  body("description", "Invalid description")
+    .trim()
+    .escape()
+    .isLength({ max: 2000 })
+    .withMessage("Description max length is 2,000 characters"),
```
lxkedinh: The description property for event creation is optional, it needs a `.optional()` check in this chain.

packages/server/validators/events.js
```diff
@@ -0,0 +1,203 @@
...
+
+  //Location checks
+  body("location", "Invalid location")
+    .trim()
+    .escape()
+    .isLength({ max: 255 })
+    .withMessage("Location max length is 255 characters"),
```
lxkedinh: The location property for event creation is optional, it needs a `.optional()` check in this chain.

packages/server/validators/events.js
```diff
@@ -0,0 +1,203 @@
...
+  body("thumbnail", "Invalid thumbnail")
+    .trim()
+    .escape()
+    .isLength({ max: 255 })
+    .withMessage("Thumbnail max length is 255 characters")
+    .custom(async (value) => {
+      if (!/\.(jpg|png|JPG|PNG|JPEG|jpeg)$/.test(value)) {
+        throw new Error("Thumbnail is not a valid image file");
+      }
+    }),
```
lxkedinh: The thumbnail property for event creation is optional, it needs a `.optional()` check in this chain.

packages/server/validators/events.js
```diff
@@ -0,0 +1,203 @@
...
+
+  //Start Date checks
+  //Convert end date to a Date object for easier comparison
+  check("endDate").toDate(),
```
lxkedinh: `check("endDate").toDate()` isn't correct here, you already have a validator for the end date input field below on line 83. Get rid of this `check()` along with the one on line 82. I assume you're doing this to make sure that the other input date field is present before validating. You should use `.if()` for conditional validating, only continue validating if the other date field is present. Check [here](https://express-validator.github.io/docs/api/validation-chain/#if).

packages/server/validators/events.js
```diff
@@ -0,0 +1,203 @@
...
+    .toDate()
+    .withMessage("Start date is not in a valid date format")
+    .custom(async (startDate, { req }) => {
+      //If new end date is provided, check if start date is before the end date
+      if (req.body.endDate) {
+        if (startDate > req.body.endDate) {
+          throw new Error("Start date is after new end date");
+        }
+      }
+    }),
```
lxkedinh: Start date input field is optional, add a `.optional()` check to it.

packages/server/validators/events.js
```diff
@@ -0,0 +1,203 @@
...
+    .isISO8601()
+    .toDate()
+    .withMessage("End date is not in a valid date format")
+    .custom(async (endDate, { req }) => {
+      if (req.body.startDate) {
+        if (endDate < req.body.startDate) {
+          throw new Error("End date is before new start date");
+        }
+      }
+    }),
```
lxkedinh: End date input field is optional, add a `.optional()` check to it.

packages/server/validators/events.js
lxkedinh: Regarding `startDate` and `endDate`, we only allow two options for these input fields.
1. None of them are present.
2. Both of them are present and both represent a proper date range.

So, we want to raise errors when
- only one of the fields is present
- the end date is before the start date.

You'll have to modify the custom validators on both `startDate` and `endDate` a little.

Add a guard clause to throw an error if the other date field is not present.
```js
// startDate custom validator
.custom(async (startDate, { req }) => {
  if (!req.body.endDate) {
    throw new Error("Cannot have a start date without an end date");
  }
})

// endDate custom validator
.custom(async (endDate, { req }) => {
  if (!req.body.startDate) {
    throw new Error("Cannot have an end date without a start date");
  }

  // only the endDate validator needs the check to make sure the date is a proper range
  if (endDate < req.body.startDate) {
    throw new Error("End date cannot be before start date");
  }
})
```

packages/server/validators/events.js
```diff
@@ -0,0 +1,203 @@
...
+      if (req.body.endDate) {
+        if (startDate > req.body.endDate) {
+          throw new Error("Start date is after new end date");
+        }
+      }
+    }),
+
+  //End Date checks
+  //Similar checks to start sate but makes sure end date is after the start
+  check("startDate").toDate(),
```
lxkedinh: `check("startDate").toDate()` isn't correct here, same thing as my earlier comment with `check("endDate").toDate()`. Check [here](https://express-validator.github.io/docs/api/validation-chain/#if) for conditional validating.

packages/server/validators/events.js
```diff
@@ -0,0 +1,203 @@
...
+    .isLength({ max: 255 })
+    .withMessage("Thumbnail max length is 255 characters")
+    .custom(async (value) => {
+      if (!/\.(jpg|png|JPG|PNG|JPEG|jpeg)$/.test(value)) {
+        throw new Error("Thumbnail is not a valid image file");
+      }
+    }),
+  //Start Date checks
+  //Convert end date to a Date object for easier comparison
+  check("endDate").optional().toDate(),
```
lxkedinh: Same thing as my earlier comment regarding conditional validating. See [here](https://express-validator.github.io/docs/api/validation-chain/#if).

packages/server/validators/events.js
```diff
@@ -0,0 +1,203 @@
...
+        //Checks if new start date is before kept end date
+        const currEvent = await EventController.getEvent(req.params.eventId);
+        if (currEvent.endDate != null && startDate > currEvent.endDate) {
+          throw new Error("Start date is after end date");
+        }
+      }
+    }),
+  //End Date checks
+  //Similar checks to start sate but makes sure end date is after the start
+  check("startDate").optional().toDate(),
```
lxkedinh: Same thing as my earlier comment regarding conditional validating. See [here](https://express-validator.github.io/docs/api/validation-chain/#if).

packages/server/validators/events.js
```diff
@@ -0,0 +1,203 @@
...
+    .withMessage("Not a UUID")
+    .exists({ checkFalsy: true })
+    .withMessage("Event Id cannot be null or empty")
+    .custom(async (value) => {
+      try {
+        await EventController.getEvent(value); //Check if event exists by getting it
+      } catch (error) {
+        //If event does not exists, then a NotFoundError is thrown from controller
+        throw new Error(error);
+      }
```
lxkedinh: Don't throw the actual Prisma error, this sends the full Prisma error message back to the frontend which isn't helpful. Replace `throw new Error(error);` with something like
```js
throw new Error(`No event exists with an ID of ${value}`);
```

packages/server/validators/events.js
```diff
@@ -0,0 +1,203 @@
...
+          throw new Error("Start date is after new end date");
+        }
+      } else {
+        //Checks if new start date is before kept end date
+        const currEvent = await EventController.getEvent(req.params.eventId);
+        if (currEvent.endDate != null && startDate > currEvent.endDate) {
+          throw new Error("Start date is after end date");
+        }
+      }
+    }),
```
lxkedinh: The start and end date input fields for updating events are a bit more complicated to validate. You'll have to modify both the `startDate` and `endDate` custom validators to cover all these.

We need to make sure it covers all the following cases:
- new provided start date, but current end date is null (invalid)
- new provided end date, but current start date is null (invalid)
- new provided start date, but after current end date (invalid)
- new provided end date, but before current start date (invalid)
- new provided start and end date, but not proper date range because end is before start date (invalid)

packages/server/validators/events.js
```diff
@@ -0,0 +1,203 @@
...
+        if (endDate < req.body.startDate) {
+          throw new Error("End date is before new start date");
+        }
+      } else {
+        const currEvent = await EventController.getEvent(req.params.eventId);
+        if (currEvent.startDate != null && endDate < currEvent.startDate) {
+          throw new Error("End date is before start date");
+        }
+      }
+    }),
```
lxkedinh: Same thing I said in an earlier comment regarding the updating event validators for `startDate` and `endDate`. You'll have to modify this to cover all the cases I listed in my earlier comment.

packages/server/routes/api/events.js
```diff
@@ -114,4 +171,84 @@ router.get(
...
+    if (!result.isEmpty()) {
+      response.status(400).json({
+        status: "fail",
+        data: result.array(),
+      });
+      return;
+    }
+
+    try {
+      const { eventId } = request.params;
```
lxkedinh: Use `matchedData(request)` to access input fields from now on since we're using express-validator. See docs for that [here](https://express-validator.github.io/docs/guides/getting-started#accessing-validated-data).

lxkedinh (changes requested):

packages/server/validators/events.js
BLiu456: We used the checks to actually convert the dates from strings to date objects. Earlier we were having trouble accessing and converting the data but I think we found an alternate solution that lets us avoid using check() for that. In the case of updating the start date, I found using the if() for conditional validation was not helpful. When the if() evaluates to false it would ignore any validations after it, effectively ending the validation early for that value. Since I needed to run certain checks based on if the end date was present, I found doing a simple if-else statement in the custom validator to be a better solution. For updating the end date, I also followed that solution to keep them consistent with each other.

packages/server/routes/api/events.js
```diff
@@ -188,7 +188,8 @@ router.delete(
     }
 
     try {
-      const { eventId } = request.params;
+      const validatedData = matchedData(request);
+      const { eventId } = validatedData.eventId;
```
lxkedinh: This should just be
```js
const eventId = validatedData.eventId;
```

The `{ }` syntax for variable initialization only works as a shorthand for extracting properties from objects.
```js
// shorthand for object.property, extract it into a variable called "property"
const { property } = object;
```

If you want to use object restructuring for this, the correct initialization would be
```js
// validatedData is an object, extract validatedData.eventId into a variable "eventId"
const { eventId } = validatedData;
```

packages/server/validators/events.js
```diff
@@ -37,60 +35,62 @@ const createEventValidator = [
...
+      //End date given but no start date
+      if (!req.body.startDate) {
+        throw new Error("Cannot have end date without start date");
+      }
+
+      //Checks if both start and end date are given
+      let startDate = new Date(req.body.startDate);
+      if (endDate < startDate) {
+        throw new Error("End date cannot be before start date");
       }
```
lxkedinh: You don't need to create a new Date object from `req.body.startDate`. In the `startDate` validator chain, the `.toDate()` method you used automatically sanitizes and converts `req.body.startDate` into a Date object for us. So, these lines should be
```js
if (endDate < req.body.startDate) {
  throw new Error("End date cannot be before start date");
}
```

packages/server/validators/events.js
```diff
@@ -130,32 +130,38 @@ const updateEventValidator = [
...
+      //Check for when new start date is given but no end date
+      else {
         const currEvent = await EventController.getEvent(req.params.eventId);
-        if (currEvent.endDate != null && startDate > currEvent.endDate) {
-          throw new Error("Start date is after end date");
+
+        //Check if the current end date is null
+        if (currEvent.endDate == null) {
+          throw new Error("Start date updated but end date is null");
+        }
```
lxkedinh: This is a nitpick but this error message implies we updated the event's start date when we didn't, the request failed. Change this error message to something more like "Cannot update start date if current end date is null".

packages/server/validators/events.js
```diff
@@ -164,16 +170,20 @@ const updateEventValidator = [
...
+      //Check for when new end date is given but no start date
+      if (!req.body.startDate) {
         const currEvent = await EventController.getEvent(req.params.eventId);
-        if (currEvent.startDate != null && endDate < currEvent.startDate) {
-          throw new Error("End date is before start date");
+
+        //Check if current start date is null
+        if (currEvent.startDate == null) {
+          throw new Error("End date updated but start date is null");
+        }
```
lxkedinh: Same thing as my earlier message about the update start date error message. This error message implies we updated the event's end date when we didn't, the request failed. Change this error message to something more like "Cannot update end date if current start date is null".

packages/server/validators/events.js
```diff
@@ -37,60 +35,62 @@ const createEventValidator = [
...
   body("thumbnail", "Invalid thumbnail")
     .trim()
     .escape()
+    .optional()
     .isLength({ max: 255 })
     .withMessage("Thumbnail max length is 255 characters")
     .custom(async (value) => {
       if (!/\.(jpg|png|JPG|PNG|JPEG|jpeg)$/.test(value)) {
         throw new Error("Thumbnail is not a valid image file");
       }
```
lxkedinh: I tested this regex on [regex101](https://regex101.com/) and it's incorrect. It's matching only the file extension but not including the file name as well so you'll need a wildcard character in this regex for that. Modify all the other event thumbnail regex checks for this as well.

lxkedinh: You don't need to use a custom validator for regex, express-validator comes with one for regex built in called `.matches()`.
```js
.matches(/regex here/),
```

lxkedinh (changes requested):
Looks really good guys! The overall functionality of these API routes is there, just a few more minor changes and then we can merge after this review.

lxkedinh (approved):

PR #35, opened 2023-12-20
Fuzzy guild search

shotoh:
- implement GET call route for fuzzy searching of guilds by name using postgresql trigrams
- minimum similarity score is 0.3

### DIFF
--- a/packages/server/controllers/guilds.js
+++ b/packages/server/controllers/guilds.js
@@ -18,7 +18,7 @@ async function getGuild(guildId) {
 
 async function searchGuild(pattern) {
   return prisma.$queryRaw`
-    SELECT * FROM guilds 
+    SELECT guild_id, name, handler, icon FROM guilds 
     WHERE SIMILARITY(name, ${pattern}) > ${MINIMUM_SIMILARITY}
     ORDER BY SIMILARITY(name, ${pattern}) DESC;`;
 }
--- a/packages/server/controllers/guilds.js
+++ b/packages/server/controllers/guilds.js
@@ -2,11 +2,6 @@ const prisma = require("../prisma/prisma");
 
 const MINIMUM_SIMILARITY = 0.3;
 
-async function getAllGuilds() {
-  const query = await prisma.guilds.findMany();
-  return query;
-}
-
 async function getGuild(guildId) {
   const query = await prisma.guilds.findFirst({
     where: {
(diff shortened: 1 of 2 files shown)


packages/server/routes/api/guilds.js
```diff
@@ -5,19 +5,45 @@ const GuildController = require("../../controllers/guilds");
...
+          guilds,
+        },
+      });
+    } catch (error) {
+      response.status(500).json({
+        status: "error",
+        message: error.message,
+      });
+    }
+  } else {
```
lxkedinh: Let's get rid of this and keep the root route to fuzzy searching for now. When we reach production, I don't see a scenario where we want to display all the guilds at once to the user.

lxkedinh (changes requested):
Code looks really good and works well so far from my testing! I just made one formatting change and broke the raw Prisma query into several lines for readability.

packages/server/controllers/guilds.js
```diff
@@ -14,7 +16,15 @@ async function getGuild(guildId) {
   return query;
 }
 
+async function searchGuild(pattern) {
+  return prisma.$queryRaw`
+    SELECT * FROM guilds 
+    WHERE SIMILARITY(name, ${pattern}) > ${MINIMUM_SIMILARITY}
+    ORDER BY SIMILARITY(name, ${pattern}) DESC;`;
+}
+
```
lxkedinh: Since these guilds will be displayed in tiny cards as items in a list for search results, we should only select what's necessary to keep the API response data size small.

Keep the fetched guild data to:
- avatar icon
- name
- handler
- guild ID

lxkedinh (changes requested):

packages/server/routes/api/guilds.js
```diff
@@ -5,14 +5,40 @@ const GuildController = require("../../controllers/guilds");
...
+    } else {
+      guilds = await GuildController.searchGuildByName(search);
+    }
+    if (guilds.length === 0) {
+      response.status(404).json({
+        status: "fail",
+        data: {
+          search: "A guild was not found in the query.",
+        },
+      });
```
lxkedinh: A more descriptive failure message would be good like `"A guild was not found with the query ${search}"`.

packages/server/controllers/guilds.js
lxkedinh, line 1: 0.3 minimum similarity looks good right now, but we should add more guilds to test out the right number for it.

lxkedinh (approved):
Looks good overall to me, no major changes needed.

packages/server/controllers/guilds.js
shotoh: the default threshold is 0.3, but we can definitely change it when needed

PR #36, opened 2024-01-08
Group Screen About Tab

ethan-ngo:
Includes:
- Description
- Links
- More Info

The data is hardcoded in. Depending on the needs of this screen, will create object props.


packages/app/screens/group/tabs/AboutScreen.js
```diff
@@ -1,8 +1,123 @@
...
+        </View>
+        <View style={styles.mediaContainer}>
+          <Ionicons
+            name="person"
+            style={styles.mediaIconStyle}
+            size={20}
+            color="#2C2C2C"
+          />
+          <Text>1,000 Members</Text>
+        </View>
```
MinT-Napkin, line 75: Looks nice! I like the layout. I would say instead of hardcoding it now you can get the guild data from the guild context, you should update from main branch since the guild context was pushed to to main.

packages/app/screens/group/tabs/AboutScreen.js
```diff
@@ -1,8 +1,123 @@
...
+            name="instagram"
+            style={styles.mediaIconStyle}
+            size={20}
+          />
+          <Text
+            style={styles.url}
+            numberOfLines={1}
+            ellipsizeMode="tail"
+            onPress={() => Linking.openURL("https://instagram.com")}>
+            https://instagram.com
```
MinT-Napkin: When getting media links from the guild context, you can get a random number of links. (i.e. no links, just a twitter link, or multiple links like twitter/x, instagram, discord, linkedin, and so on).

I would suggest using some kind of list or mapping these links or however you prefer. It may be a little complicated to display icons as well but you could possibly display icons in accordance to the link's string, although I'd say that can be an optional extra feature.

packages/app/screens/group/tabs/AboutScreen.js
```diff
@@ -1,8 +1,123 @@
...
+            color="#2C2C2C"
+          />
+          <Text>Joined January 3, 2024</Text>
+        </View>
+        <View style={styles.mediaContainer}>
+          <Ionicons
+            name="person"
+            style={styles.mediaIconStyle}
+            size={20}
+            color="#2C2C2C"
```
MinT-Napkin: The Ionicons have hardcoded colors. Just replace them with GRAY variable that you already have and that should be good.

MinT-Napkin (changes requested):
I like the layout! Just needs to use guild data now instead of hardcoded information, you should update from main and handle any conflicts. You can get guild data similar to that of the MembersScreen file.

ethan-ngo:
Ready for review again.

ethan-ngo:
Ready for review

MinT-Napkin (approved):
lgtm!

A few notes, remember to update from main every so often and try not to use npm, it lead to some inconsistencies in the yarn.lock file, hence why the pr line additions and deletions skyrocket up to the 2,000's or 4,000's.

PR #37, opened 2024-01-08
Event Overview Screen

tdang2180:
**Features**
- Implemented navigation between group screen and event overview screen
- Display event overview screen when the user presses on the event card's title
- Move back to group screen when the user presses the back arrow
- Refactored group icon and made the register/going button a separate component for reusability
- Added ability to cancel registration
- Created share button 
- Fetched event data from events API using eventID

**To Do**  (not in scope of event overview task):
- Update group screen to link an eventID with each event card


packages/app/components/EventCard/EventCardText.js
```diff
@@ -19,13 +21,28 @@ const styles = StyleSheet.create({
...
+
   return (
     <View>
       <Text style={styles.smallText}>
         {timeBegin} - {timeEnd}
       </Text>
-      <Text style={styles.eventTitle}>{title}</Text>
+      <Text style={styles.eventTitle} onPress={onTitlePress}>
+        {title}
+      </Text>
```
MinT-Napkin, line 59: I would say to add some sort of feedback when the user clicks the title section, such as a quick color difference on the background of the text, such as grey when the user physically presses on the title.

packages/app/screens/group/GroupStack.js
```diff
@@ -2,15 +2,19 @@ import React from "react";
...
   return (
     <Group.Navigator
-      screenOptions={{ headerShown: false }}
+      screenOptions={{ headerShown: false, animation: "slide_from_right" }}
       initialRouteName="GroupScreen">
       <Group.Screen name="GroupScreen" component={GroupScreen} />
+      <Group.Screen
+        name="EventOverviewScreen"
+        component={EventOverviewScreen}
+      />
```
MinT-Napkin, line 17: This file is fine but I we'll need a something like this in the FeedStack as well, since events also appear there in the feed but leads to an error of undefined since EventOverviewScreen is missing from the FeedStack.

packages/app/components/EventOverview/EventOverviewText.js
```diff
@@ -0,0 +1,80 @@
...
+  );
+};
+
+export default function EventOverviewText(props) {
+  return (
+    <View>
+      <OverviewHeader groupName={props.groupName} />
+      <Text style={styles.eventTitle}>{props.title}</Text>
+      <Text style={styles.smallText}>
+        {props.timeBegin} - {props.timeEnd}
```
MinT-Napkin, line 61: The startTime appears as 2023-11-28T21:00:30.000Z - 2023-11-21T21:00:30.000Z so it can't read that well. If you can try to reformat so its a little more readable to something like 11/28/23 at 9:00 PM or November 28, 2023 at 9:00 PM instead of a direct timestamp.

MinT-Napkin (changes requested):
Looks pretty good so far. I like the layout of screen. A few changes here and there. I would combine together EventOverview, EventOverviewText, and EventOverviewScreen together into one screen, there is a bit prop drilling which should typically avoided or handled with a react context and does get a little confusing.

(edit: Forgot to mention, don't forget to replace all test data with api calls instead)

Don't forget to update from main!

packages/app/screens/group/GroupStack.js
tdang2180: Yup, I added navigation and param to differentiate between feed and group screens

packages/app/components/EventCard/EventCardText.js
tdang2180: Good point!

MinT-Napkin (approved):
lgtm!

Made a quick change to the back button in the event overview screen to make it bigger on IOS, for whatever reason it's unreachable there but fine on Android.

I'll merge the branch for you!

PR #38, opened 2024-01-10
fix: replaced similarity search with word similarity search

shotoh:
This calculates the greatest similarity between the search pattern and any substring of the guild's name instead of the entire name as one string. It makes it easier to search for guilds which have multiple words in their name.


lxkedinh (approved):
Everything looks good to me! Fuzzy search is properly working per word now instead of the entire string.

PR #39, opened 2024-01-16
API User's Guilds

JustinNgxyen:
**Goal**:
Designed to retrieve all the guilds that a user is a member of on Icebreak. The primary objective is to provide a seamless way for developers to access a user's guild memberships programmatically.

**Implementation**:
- A GET route in users.js that uses the userId to retrieve the user's guilds using the GuildController function getGuildsForUser.
- If the userId is undefined, either not inputted or invalud, than a fail status is sent along with the reasoning. 

[RFC](https://doc.clickup.com/9011073139/d/h/8chm43k-3151/c1a634455bb0223)

### DIFF
--- a/packages/server/controllers/guilds.js
+++ b/packages/server/controllers/guilds.js
@@ -82,6 +82,41 @@ async function getGuildMembers(guildId) {
   return members;
 }
 
+async function getGuildsForUser(userId) {
+  try {
+    const user = await prisma.users.findFirst({
+      where: {
+        userId: userId,
+      },
+    });
+
+    if (!user) {
+      throw new Error(`User with ID ${userId} not found`);
+    }
+
+    const userGuilds = await prisma.guildMembers.findMany({
+      where: {
+        userId: userId,
+      },
+      include: {
+        guilds: {
+          select: {
+            icon: true,
+            guildId: true,
+            name: true,
+            handler: true,
+          },
+        },
+      },
+    });
+
+    const guilds = userGuilds.map((userGuild) => userGuild.guilds);
+    return guilds;
+  } catch (error) {
+    throw new Error(`Error fetching guilds for user: ${error.message}`);
+  }
+}
+
 module.exports = {
   getGuild,
   searchGuildByName,
@@ -92,4 +127,5 @@ module.exports = {
   deleteGuild,
   getGuildMembers,
   guildExists,
+  getGuildsForUser,
 };
--- a/packages/server/controllers/guilds.js
+++ b/packages/server/controllers/guilds.js
@@ -101,7 +101,7 @@ async function getGuildsForUser(userId) {
       include: {
         guilds: {
           select: {
-            icon: true,
+            avatar: true,
             guildId: true,
             name: true,
             handler: true,
--- a/packages/server/controllers/guilds.js
+++ b/packages/server/controllers/guilds.js
@@ -82,41 +82,6 @@ async function getGuildMembers(guildId) {
   return members;
 }
 
-async function getGuildsForUser(userId) {
-  try {
-    const user = await prisma.users.findFirst({
-      where: {
-        userId: userId,
-      },
-    });
-
-    if (!user) {
-      throw new Error(`User with ID ${userId} not found`);
-    }
-
-    const userGuilds = await prisma.guildMembers.findMany({
-      where: {
-        userId: userId,
-      },
-      include: {
-        guilds: {
-          select: {
-            avatar: true,
-            guildId: true,
-            name: true,
-            handler: true,
-          },
-        },
-      },
-    });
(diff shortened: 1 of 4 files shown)


packages/server/controllers/guilds.js
```diff
@@ -82,6 +82,41 @@ async function getGuildMembers(guildId) {
...
+          },
+        },
+      },
+    });
+
+    const guilds = userGuilds.map((userGuild) => userGuild.guilds);
+    return guilds;
+  } catch (error) {
+    throw new Error(`Error fetching guilds for user: ${error.message}`);
+  }
```
lxkedinh: Get rid of this catch block since the route handler already has a catch block to catch this error if it occurs.

packages/server/controllers/guilds.js
```diff
@@ -82,6 +82,41 @@ async function getGuildMembers(guildId) {
...
+  try {
+    const user = await prisma.users.findFirst({
+      where: {
+        userId: userId,
+      },
+    });
+
+    if (!user) {
+      throw new Error(`User with ID ${userId} not found`);
+    }
```
lxkedinh: These checks won't be necessary. I'm getting all of us to migrate over to express-validator to streamline error handling. I'll tell you more details in Discord but you can start by reading docs [here](https://express-validator.github.io/docs/guides/getting-started#adding-a-validator).

packages/server/routes/api/users.js
```diff
@@ -53,4 +54,42 @@ router.get(
...
+      const user = await UserController.getUser(userId);
+
+      if (!user) {
+        return response.status(404).json({
+          status: "fail",
+          data: {
+            userId: `User not found with ID ${userId}`,
+          },
+        });
+      }
```
lxkedinh: Like I said in earlier comment, we'll be using [express-validator](These checks won't be necessary. I'm getting all of us to migrate over to express-validator to streamline error handling. I'll tell you more details in Discord but you can start by reading docs [here](https://express-validator.github.io/docs/guides/getting-started#adding-a-validator).) for error handling now. Eventually, these checks will be replaced by an express validator instead.

packages/server/controllers/guilds.js
```diff
@@ -82,6 +82,41 @@ async function getGuildMembers(guildId) {
   return members;
 }
 
+async function getGuildsForUser(userId) {
```
lxkedinh: Since the API route for this is located in `api/users.js`, let's move this function to `controllers/users.js` for consistency.

lxkedinh (changes requested):

packages/server/routes/api/users.js
```diff
@@ -58,24 +58,36 @@ router.get(
...
-
-      // Check if the user exists
-      const user = await UserController.getUser(userId);
+    const result = validationResult(request);
+    if (!result.isEmpty()) {
+      return response.status(400).json({
+        status: "fail",
+        data: {
+          errors: result.array(),
+        },
```
lxkedinh: You don't need to enclose the `result.array()` in another object.
```js
return response.status(400).json({
  status: "fail",
  data: result.array()
});
```

packages/server/routes/api/users.js
```diff
@@ -58,24 +58,36 @@ router.get(
...
+      .isUUID()
+      .withMessage("Invalid user ID provided")
+      .custom(async (value) => {
+        const user = await UserController.getUser(value);
+        if (!user) {
+          throw new Error("User with ID ${value} not found");
+        }
+        return true;
+      }),
+  ],
```
lxkedinh: We don't want to clutter the route handler files with all the validators for our routes, so move this to `packages/server/validators/users.js` and export it to use here instead. You can look at the guild validators in main branch for reference on how to structure the file.

lxkedinh (changes requested):

packages/server/validators/users.js
```diff
@@ -0,0 +1,23 @@
...
+    .blacklist("<>")
+    .isUUID()
+    .withMessage("Invalid user ID provided")
+    .custom(async (value) => {
+      const user = await UserController.getUser(value);
+      if (!user) {
+        throw new Error("User with ID ${value} not found");
+      }
+      return true;
+    }),
```
lxkedinh: This custom validator shouldn't return anything since this is not the correct use case for that.

From the docs,
```
Custom validators can be asynchronous, in which case it can return a promise.
The returned promise is awaited on, and it must resolve in order for the field to be valid.
If it rejects, the field is deemed invalid.
```

lxkedinh (approved):
I'll approve this PR now but check my last comment for one last minor change that needs to be made before you merge. Make sure your branch is up to date with main as well before squashing and merging.

PR #40, opened 2024-01-26
Express Validators - Guild CRUD API Routes 

larrylaa:
Refactored old error handling into event validation using express validator for the following routes.

- Get Guild
- Create Guild
- Update Guild
- Delete Guild


package.json
lxkedinh, line 1: You installed `express-validator` into the wrong package.json. This is a monorepo so it's split up between the frontend and backend repos. Remove `express-validator` on the root package.json and change directory into `packages/server` and install `express-validator` in there instead. This is the general rule of thumb you should follow, too. If you want to install dependencies, you should first make sure you're in `packages/server` to install dependencies for the backend only.

packages/server/routes/api/guilds.js
```diff
@@ -57,37 +59,29 @@ router.get("/", AuthController.authenticate, async (request, response) => {
...
   AuthController.authenticate,
+  guildIdValidator,
   async (request, response) => {
+    const result = validationResult(request);
+
+    if (!result.isEmpty()) {
+      return response.status(400).json({
+        status: "fail",
+        data: result.array({ onlyFirstError: true }),
+      });
```
lxkedinh: For the express-validator error handling on all your routes, change these to just `result.array()` to return all the validation errors.

packages/server/validators/guilds.js
```diff
@@ -0,0 +1,219 @@
...
+];
+
+const createGuildValidator = [
+  // Name Check
+  body("name", "Invalid guild name.")
+    .exists({ checkFalsy: true })
+    .withMessage("Name can't be null or empty.")
+    .trim()
+    .isLength({ min: 1, max: 255 })
+    .withMessage("Name length must be between 1 to 255 characters."),
```
lxkedinh: Guild names are max 100 characters. This applies to your update guild validator too.

packages/server/validators/guilds.js
```diff
@@ -0,0 +1,219 @@
...
+
+  // Guild Handler Name Check
+  body("handler", "Invalid guild handler name.")
+    .exists({ checkFalsy: true })
+    .withMessage("Guild handler's name can't be null or empty.")
+    .trim()
+    .isLength({ min: 1, max: 255 })
+    .withMessage("Guild handler's name must be between 1 to 255 characters."),
```
lxkedinh: Guild handlers are max 50 characters. This applies to your update guild validator too.

packages/server/validators/guilds.js
lxkedinh, line 1: You should add `.escape()` to all your validator checks if they don't have them. It's generally a good idea to escape all input to prevent any possible code injection and to sanitize all the input.

packages/server/validators/guilds.js
```diff
@@ -0,0 +1,219 @@
...
+  body("website", "Invalid guild website.")
+    .optional()
+    .trim()
+    .isLength({ max: 2048 })
+    .withMessage("Guild thumbnail max length is 2048 characters.")
+    .matches(/^(https?|ftp):\/\/[^\s/$.?#].[^\s]*$/)
+    .withMessage("Provided website url is invalid."),
+
+  // Tags Check
+  body("tags", "Invalid tags.").isArray().withMessage("Tags must be an array."),
```
lxkedinh: Guild tags are optional. This applies to your update guild validator too.

packages/server/validators/guilds.js
```diff
@@ -0,0 +1,219 @@
...
+    .escape()
+    .isUUID()
+    .withMessage("Guild ID not a valid UUID.")
+    .custom(async (value) => {
+      try {
+        await GuildController.getGuild(value);
+      } catch (error) {
+        throw new Error(`No guild exists with an ID of ${value}`);
+      }
+    }),
```
lxkedinh: You can get rid of this duplicate `guildId` parameter field check since you already have this in `guildIdValidator`. You can use more than one validator in an express API route and they'll run in order of definition like so.

```js
router.put(
  "/:guildId",
  AuthController.authenticate,
  // runs guildIdValidator then updateGuildValidator
  guildIdValidator,
  updateGuildValidator,
  // ...
});
```

lxkedinh (changes requested):

packages/server/validators/guilds.js
larrylaa: Omitting .escape().

lxkedinh, line 1: Let's go with your method of using `.blacklist()` for `<` and `>` characters instead of using `.escape()` for the validators.

lxkedinh (changes requested):

lxkedinh (approved):
Everything looks good to me! I'll merge this PR for you.

PR #41, opened 2024-01-27
feat: event attendees route validator (#41)

HStevenC:
Refactored the API route to get the attendee users of an event to use express-validator.


packages/server/routes/api/events.js
```diff
@@ -117,26 +117,23 @@ router.get(
...
+      .blacklist("<>")
+      .isUUID()
+      .withMessage((eventId) => `Given eventId ${eventId} is invalid.`)
+      .bail()
+      .custom(async (eventId) => {
+        if (!(await EventController.existsInPrisma(eventId))) {
+          throw new Error(`No event exists with an ID of ${eventId}`);
+        }
+      }),
+  ],
```
lxkedinh: Since we will have many validators for the different routes, it would be best to put the validators in their own file so we don't clutter up the route handler files. The convention the rest of the team is doing is putting validators into a `packages/server/validators` folder and since you're doing events it'll be in `packages/server/validators/events.js`. Move this validator into that file and then export it to use here instead.

packages/server/routes/api/events.js
```diff
@@ -117,26 +117,23 @@ router.get(
...
-      response.status(404).json({
-        status: "fail",
-        data: {
-          eventId: `No event exists with an ID of ${eventId}`,
-        },
-      });
-      return;
+    const errors = validationResult(request);
+    if (!errors.isEmpty()) {
+      return response.status(400).json({ errors: errors.array() });
```
lxkedinh: Remember to follow our API response convention, this would be a fail status response.
```js
response.status(400).json({
  status: "fail",
  data: errors.array()
});
```

packages/server/routes/api/events.js
```diff
@@ -117,26 +117,23 @@ router.get(
...
     }
     try {
       const { eventId } = request.params;
```
lxkedinh: Since we're using express-validator now, we need to import `matchedData()` to get the validated data from express-validator if it passes. Don't use `request.params` and `request.body` anymore.

```js
const { eventId } = matchedData(request);
```

lxkedinh (changes requested):

lxkedinh (approved):

PR #42, opened 2024-02-09
Justin/api guild leaderboard

JustinNgxyen:
This API route will allow the app to display a leaderboard for the guild and the top participating members. This is our way of "gamifying" the app and incentivizing participation for university clubs in our app.

API route should return

- list of guild members sorted by how many points they have from most to least
- the following guild member info
- first and last name
- avatar
- their current amount of points in the guild

### DIFF
--- a/packages/server/controllers/guilds.js
+++ b/packages/server/controllers/guilds.js
@@ -82,35 +82,34 @@ async function getGuildMembers(guildId) {
   return members;
 }
 
-async function getGuildMembersPoints(guildId) {
-  const getMembers = await prisma.guildMembers.findMany({
+async function guildLeaderboard(guildId) {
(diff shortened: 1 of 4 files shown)


packages/server/controllers/guilds.js
```diff
@@ -82,6 +82,38 @@ async function getGuildMembers(guildId) {
...
+          handler: true,
+          avatar: true,
+        },
+      },
+    },
+
+    orderBy: [
+      {
+        points: "asc",
+      },
```
lxkedinh: Change this to order by descending. We want to sort by guild members with most points to least points.

lxkedinh (changes requested):

lxkedinh (approved):

PR #43, opened 2024-02-14
Users guild list screen

MichaelWuhu:


packages/app/screens/group/GuildCard.js
```diff
@@ -0,0 +1,86 @@
...
+  return (
+    <TouchableOpacity style={styles.card} onPress={() => openGuild(props.id)}>
+      <GroupIcon
+        style={styles.avatar}
+        testID="groupIcon"
+        icon={
+          props.avatar
+          // ? props.avatar
+          // : require("@app/assets/test-club-icon.png")
+        } // TEMPORARY USING test-club-icon ... switch out with something else later for n/a icon
```
MinT-Napkin: Don't forget to remove any commented out code / console logs since we don't want that in production code!

packages/app/screens/group/GuildCard.js
```diff
@@ -0,0 +1,86 @@
...
+import React from "react";
+import { View, Text, StyleSheet } from "react-native";
+import PropTypes from "prop-types";
+import { TouchableOpacity } from "react-native";
+
+const GuildCard = (props) => {
+  const iconSize = 50;
+
+  const openGuild = (groupId) => {
+    props.navigation.navigate("GroupScreen", { groupId: groupId });
```
MinT-Napkin: Change variable name groupId to guildId, as GuildProvider in GroupScreen is expecting a guildId not a groupId.

MinT-Napkin (changes requested):
Looks pretty good so far! I love the look of the guild list.

Although there is an issue where if the user clicks on a guild, it doesn't show up any of the correct info from that guild.

Though I think changing the variable name should do the trick, note that Cal Poly Swift and CPP Computer Science Society guilds in the test user wouldn't actually work and give a error code 400 because they don't have existing banners/icons images, I've added a guild with that, 'Fumo' is the name, to the test user. You can alternate between SEA guild and Fumo guild to test if it works.

Once that is working I believe this pr is ready to be pushed to main, @ me on Discord once again to review changes as soon as I can 👍.

MinT-Napkin (approved):
lgtm!

PR #44, opened 2024-02-17
API Checking in to Guild Event

JustinNgxyen:
When an event is about to start or has started recently, we want to allow event attendees to check in to the event so they're eligible to gain points for club participation. Guild members can only obtain points if they have checked in. 

This API route will be called to change the status of a guild member:
- not interested
- interested
- attending
- checked in

[RFC](https://doc.clickup.com/9011073139/d/h/8chm43k-3571/29ca28344f7005b)

### DIFF
--- a/packages/server/controllers/events.js
+++ b/packages/server/controllers/events.js
@@ -122,6 +122,22 @@ async function getEventAttendees(eventId) {
   return query;
 }
 
+async function updateAttendeeStatus(eventId, userId, attendeeStatus) {
+  const query = await prisma.eventAttendees.update({
+    where: {
+      userId_eventId: {
+        userId: userId,
+        eventId: eventId,
+      },
+    },
+    data: {
+      status: attendeeStatus,
+    },
+  });
+
+  return query;
+}
+
 module.exports = {
   getEvent,
   getEvents,
@@ -131,4 +147,5 @@ module.exports = {
   updateEvent,
   createEvent,
   getEventAttendees,
+  updateAttendeeStatus,
 };
--- a/packages/server/controllers/events.js
+++ b/packages/server/controllers/events.js
@@ -121,6 +121,73 @@ async function getEventAttendees(eventId) {
   return query;
 }
 
+async function updateAttendeeStatus(eventId, userId, attendeeStatus) {
+  const query = await prisma.eventAttendees.update({
+    where: {
+      userId_eventId: {
+        userId: userId,
+        eventId: eventId,
+      },
+    },
+    data: {
+      status: attendeeStatus,
+    },
+  });
+
+  if (attendeeStatus === "CheckedIn") {
+    await addCheckInPoints(eventId, userId);
+  }
+
+  return query;
+}
+
+async function addCheckInPoints(eventId, userId) {
+  const event = await prisma.events.findUnique({
+    where: {
+      eventId: eventId,
+    },
+    select: {
+      startDate: true,
+      guildId: true, // Assuming you need the guild ID to find the guildMember
+    },
+  });
+
+  if (!event) throw new Error("Event not found");
+
+  const currentTime = new Date();
+  const eventStartTime = event.startDate.getTime();
+  const fiveMinutesInMilliseconds = 5 * 60 * 1000;
+  const currentTimeInMilliseconds = currentTime.getTime();
+  //Default points to add
+  let pointsToAdd = 3;
+
+  if (
+    currentTimeInMilliseconds <= eventStartTime &&
+    currentTimeInMilliseconds >= eventStartTime - fiveMinutesInMilliseconds
+  ) {
+    pointsToAdd = 5;
+  } else if (
+    currentTimeInMilliseconds > eventStartTime &&
+    currentTimeInMilliseconds <= eventStartTime + fiveMinutesInMilliseconds
+  ) {
+    pointsToAdd = 4;
+  }
+
+  await prisma.guildMembers.update({
+    where: {
+      userId_guildId: {
+        userId: userId,
+        guildId: event.guildId,
+      },
+    },
+    data: {
+      points: {
+        increment: pointsToAdd,
+      },
+    },
+  });
+}
+
 module.exports = {
   getEvent,
   getEvents,
@@ -130,4 +197,5 @@ module.exports = {
   updateEvent,
   createEvent,
   getEventAttendees,
+  updateAttendeeStatus,
 };
--- a/packages/server/controllers/events.js
+++ b/packages/server/controllers/events.js
@@ -122,7 +122,7 @@ async function getEventAttendees(eventId) {
   return query;
 }
 
-async function updateAttendeeStatus(eventId, userId, attendeeStatus) {
+async function updateAttendeeStatus(eventId, userId) {
   const query = await prisma.eventAttendees.update({
     where: {
       userId_eventId: {
@@ -131,7 +131,7 @@ async function updateAttendeeStatus(eventId, userId, attendeeStatus) {
       },
     },
     data: {
-      status: attendeeStatus,
+      status: "CheckedIn",
     },
   });
 
--- a/packages/server/controllers/events.js
+++ b/packages/server/controllers/events.js
@@ -122,14 +122,19 @@ async function getEventAttendees(eventId) {
 }
 
 async function updateAttendeeStatus(eventId, userId, attendeeStatus) {
-  const query = await prisma.eventAttendees.update({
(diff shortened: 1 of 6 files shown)


packages/server/routes/api/events.js
```diff
@@ -280,4 +281,58 @@ router.get(
...
+      response.status(400).json({
+        status: "fail",
+        data: errors.array(),
+      });
+      return;
+    }
+
+    try {
+      const { status, userId } = request.body;
+      const eventId = request.params.eventId;
```
lxkedinh: Remember that we shouldn't access request body and parameters like this anymore now that we have `express-validator`. Import and use `matchedData(request)` from `express-validator`.

```js
const { userId, eventId } = matchedData(request);
```

packages/server/validators/users.js
```diff
@@ -18,6 +18,23 @@ const userIdValidator = [
     }),
 ];
 
+const userIdBodyValidator = [
+  body("userId")
+    .notEmpty()
+    .withMessage("User ID cannot be empty")
+    .blacklist("<>")
+    .isUUID()
+    .withMessage("Invalid user ID provided")
```
lxkedinh, line 27: Add a `.bail()` check to this UUID check so it looks like this.
```js
  //...
  .isUUID()
  .withMessage("Invalid user ID provided")
  .bail()
  // .custom( ....
```

Adding `.bail()` makes it so the validator terminates immediately after failing any previous checks instead of continuing with validation. This is useful to stop validation early before hitting our database for the custom validator which helps with performance.

image

packages/server/routes/api/events.js
```diff
@@ -280,4 +281,58 @@ router.get(
...
+      const event = await EventController.getEvent(eventId);
+
+      const currentTime = new Date();
+      const checkInStartTime = new Date(event.startDate.getTime() - 5 * 60000);
+      const checkInEndTime = new Date(event.endDate.getTime() + 15 * 60000);
```
lxkedinh: You should stray away from using "magic numbers", numbers that are directly in code. It makes the code less readable because if someone else reads this or if you come back to it again in the future, it's not immediately clear what these numbers mean.

To fix this, assign the numbers to variables with clear, meaningful names instead and then use those here.

You can read more about magic numbers [here](https://stackoverflow.com/questions/47882/what-are-magic-numbers-and-why-do-some-consider-them-bad).

packages/server/routes/api/events.js
```diff
@@ -280,4 +281,58 @@ router.get(
...
+    if (!errors.isEmpty()) {
+      response.status(400).json({
+        status: "fail",
+        data: errors.array(),
+      });
+      return;
+    }
+
+    try {
+      const { status, userId } = request.body;
```
lxkedinh: This route is just for checking in users to events, so we should get rid of the `status` request body parameter since the status will always be `CheckedIn` when updating the event attendee status in the events controller.

lxkedinh (changes requested):

packages/server/controllers/events.js
lxkedinh, line 1: GitHub's not letting me select specific lines to comment on for this file for some reason, so just typing it here. Revert your change to `updateAttendeeStatus()`, it should still have the `attendeeStatus` third parameter. In my previous comment where I said `status` should always be `CheckedIn`, I meant to pass it as a parameter in your route handler.
```js
const eventAttendeeData = await EventController.updateAttendeeStatus(
  eventId,
  userId,
  "CheckedIn"
);
```

packages/server/routes/api/events.js
```diff
@@ -298,19 +298,21 @@ router.post(
...
-      const eventId = request.params.eventId;
+      const { userId, eventId } = matchedData(request);
       const event = await EventController.getEvent(eventId);
 
       const currentTime = new Date();
-      const checkInStartTime = new Date(event.startDate.getTime() - 5 * 60000);
-      const checkInEndTime = new Date(event.endDate.getTime() + 15 * 60000);
+
+      const EVENT_START_TIME = event.startDate.getTime() - 5 * 60000;
+      const EVENT_END_TIME = event.endDate.getTime() + 15 * 60000;
```
lxkedinh:
```js
const EVENT_END_TIME = event.endDate.getTime() + 15 * 60000;
```
The above line is incorrect. How it is right now means 15 minutes after the event has ended. We want to stop allowing check in 15 minutes after the event has started so it should be
```js
const EVENT_END_TIME = event.startDate.getTime() + 15 * 60000;
```


The 60000 is still a magic number, so put it into a variable to clarify what exactly it's for which in this case is for unit conversion from minutes to milliseconds.
```js
const MINUTES_TO_MILLISECONDS = 60000;
```

- change `EVENT_START_TIME` to `eventCheckInStartBound`
- change `EVENT_END_TIME` to `eventCheckInEndBound`

These would be more specific and clearer names for these variables since they represent the beginning and end of the time window we're allowing people to check in. They don't represent the actual event start and end times.

```js
const MINUTES_TO_MILLISECONDS = 60000;
const eventCheckInStartBound = event.startDate.getTime() - 5 * MINUTES_TO_MILLISECONDS;
const eventCheckInEndBound = event.startDate.getTime() + 15 * MINUTES_TO_MILLISECONDS;
```

In the code block above, after getting rid of the magic numbers and using more specific variable names, it's more clear that we're allowing event check in 5 minutes before and ending check in 15 minutes after the start time.

lxkedinh (changes requested):

lxkedinh (commented):
These changes look good. Last thing to implement is adding points on checking in and this will be finished.

packages/server/routes/api/events.js
```diff
@@ -281,40 +281,57 @@ router.get(
...
-      response.status(200).json({
-        status: "success",
-        data: {
-          upcomingEvents: upcoming,
-        },
-      });
+      const checkInStartTime = new Date(eventCheckInStartBound);
+      const checkInEndTime = new Date(eventCheckInEndBound);
+
+      if (currentTime >= checkInStartTime && currentTime <= checkInEndTime) {
```
lxkedinh: The logic to check whether the current time is within the check-in time window can be moved to an express-validator since this is part of data validation before we execute whatever the client requests, in this case, to check in a user into an event.

Moving this logic into an express-validator means you can remove the if block here for cleaner code.
```js
// express-validator validates the current time to make sure it's within check-in time window
// no need to keep this code in an if block anymore
const eventAttendeeData = await EventController.updateAttendeeStatus(
  eventId,
  userId,
  "CheckedIn"
);

response.status(200).json({
  status: "success",
  data: {
    eventRegistration: eventAttendeeData,
  },
});
```

lxkedinh (changes requested):
You also need to add a custom express-validator to check whether the requesting user is actually a part of the guild that the event is hosted under.

From the input `userId`, check if the guild with the `guildId` from the input `eventId` has a member with `userId`.

packages/server/validators/events.js
```diff
@@ -193,8 +193,44 @@ const eventIdValidator = [
...
+      const eventCheckInStartBound =
+        event.startDate.getTime() - 5 * MINUTES_TO_MILLISECONDS;
+      const eventCheckInEndBound =
+        event.startDate.getTime() + 15 * MINUTES_TO_MILLISECONDS;
+      const checkInStartTime = new Date(eventCheckInStartBound);
+      const checkInEndTime = new Date(eventCheckInEndBound);
+
+      if (currentTime < checkInStartTime || currentTime > checkInEndTime) {
+        throw new Error("Check-in window has closed.");
+      }
```
lxkedinh: We should have this error message be a bit more specific. It's weird if the frontend sees the error message "Check-in window has closed." before the event has even started. A different error message for before and after the check in time window would be better.

```js
if (currentTime < checkInStartTime) {
  throw new Error("This event has not opened for check-in yet.");
}
if (currentTime > checkInEndTime) {
  throw new Error("Check-in window has closed.");
}
```

packages/server/controllers/guilds.js
```diff
@@ -114,6 +114,16 @@ async function getLeaderboard(guildId) {
   return members;
 }
 
+async function isGuildMember(guildId, userId) {
+  const guildMember = await prisma.guildMembers.findFirst({
+    where: {
+      guildId: guildId,
+      userId: userId,
+    },
+  });
```
lxkedinh: use `findUnique()` instead of `findFirst()`, we're trying to query by the guild members table's unique identifier, the `guildId` and `userId`.

lxkedinh (changes requested):

lxkedinh (approved):

PR #47, opened 2024-03-01
Forgot password & reset API routes

larrylaa:
- Implemented API routes for forgot and reset password user workflow

**forgot password route**
- Will find the user's account from email, and send the user a reset link with an embedded JSON web token.

**reset password route**
- Will take a token and new password, reset the user's password, and send them a confirmation email.


packages/server/package.json
```diff
@@ -18,13 +18,15 @@
     "bcrypt": "^5.1.0",
     "cookie-parser": "^1.4.6",
     "cors": "^2.8.5",
+    "crypto": "^1.0.1",
```
lxkedinh: I believe `crypto` should already be part of the standard Node.js modules so you shouldn't have to install it.

https://nodejs.org/docs/latest-v16.x/api/crypto.html#crypto

packages/server/utils/token.js
```diff
@@ -30,6 +31,19 @@ function generateAccessToken(user) {
   );
 }
 
+function generateResetPasswordToken(user) {
+  const userId = user;
```
lxkedinh: Don't forget the curly braces for object destructuring to extract object properties into variables. Here you're just assigning the user object to `userId`. As you have it right now, The `userId` variable doesn't actually represent the user ID, it's holding the entire user object.

```js
const { userId } = user;
```

packages/server/controllers/auth.js
```diff
@@ -189,6 +190,100 @@ async function authenticate(request, response, next) {
...
+  const result = await prisma.users.findUnique({
+    where: {
+      email: email,
+    },
+  });
+
+  if (result === null) return false;
+
+  return true;
+}
```
lxkedinh: Since this returns a boolean like your other `isGoogleAccount()` function, change the name to start with "is" like `isUserEmail()` as well for consistency. That's also the convention for helper functions that check if something "is" _____.

packages/server/controllers/auth.js
```diff
@@ -189,6 +190,100 @@ async function authenticate(request, response, next) {
...
+      },
+    });
+
+    const info = await transporter.sendMail({
+      from: "icebreak@cppicebreak.com", // sender address
+      to: email, // list of receivers
+      subject: "Icebreak: Password Reset Request", // Subject line
+      text: `Please click the following link to reset your password: ${link}`, // plain text body
+      // html: "<b>Hello world?</b>", // html body (THE HTML email can be a frontend task.)
+    });
```
lxkedinh: If you take a closer look at when you're creating the transport in this `sendPasswordResetEmail(email, link)` function and the `sendPasswordResetConfirmationEmail(email)` function, there is duplicate code where the only things being changed are the email subject and the email text.

Generally, any duplicated code is a sign that you need to rewrite it. Good practice is to reduce duplicate code and make code as reuseable as possible. My suggestion is to write a base `sendEmail(address, subject, body)` function and then change your `sendPasswordResetEmail()` and `sendPasswordResetConfirmationEmail()` to call the new `sendEmail(address, subject, body)`.

Structuring the code this way allows for easy extension to send different emails when writing new features in the future.

```js
async function sendEmail(address, subject, body) {
  /** 
   * no try/catch because possible errors are thrown up
   * until it reaches the route handler try/catch
   */ 
  const transporter = nodemailer.createTransport({
      host: "mail.privateemail.com",
      port: 465,
      secure: true,
      auth: {
        user: "icebreak@cppicebreak.com",
        pass: process.env.MAILBOX_PASSWORD,
      },
    });

    const info = await transporter.sendMail({
      from: "icebreak@cppicebreak.com",
      to: address,
      subject: subject,
      text: body,
    });
}

/**
 * Right now, you guys pass the link to sendPasswordResetEmail
 * but you should instead pass the password reset token and form the link
 * inside the function.
 * 
 * It's better to create and keep the link contained within the function where
 * it is directly used.
 */ 
async function sendPasswordResetEmail(address, passwordResetToken) {
  const subject = "Icebreak: Password Reset Request";
  const link = `http://cppicebreak.com/reset-password?token=${passwordResetToken}`; // changing to icebreak domain for demonstration
  const body = `Please click the following link to reset your password: ${link}`;

  sendEmail(address, subject, body);
}

async function sendPasswordResetConfirmationEmail(address) {
  const subject = "Icebreak: Password Changed";
  const body = "Your Icebreak password has been changed. If you did not request a password change, please immediately contact us at icebreak@cppicebreak.com.";

  sendEmail(address, subject, body);
}
```

packages/server/routes/api/auth.js
```diff
@@ -283,4 +296,168 @@ router.post("/token/revoke", async (request, response) => {
...
+      // Check if the user has an existing account.
+      const verifyEmailResult = await AuthController.verifyUserEmail(email);
+      if (!verifyEmailResult) {
+        return response.status(400).json({
+          status: "fail",
+          data: {
+            message: "User with given email does not exist.",
+          },
+        });
+      }
```
lxkedinh: This can be moved to an express validator inside a `.custom()` check.

packages/server/routes/api/auth.js
```diff
@@ -283,4 +296,168 @@ router.post("/token/revoke", async (request, response) => {
...
+      );
+
+      if (isGoogleAccountResult) {
+        return response.status(400).json({
+          status: "fail",
+          data: {
+            message: "Please login using Google.",
+          },
+        });
+      }
```
lxkedinh: This can be moved to an express validator inside a `.custom()` check.

packages/server/routes/api/auth.js
```diff
@@ -283,4 +296,168 @@ router.post("/token/revoke", async (request, response) => {
...
+
+      // Check if the JWT token has been used before (check if it's in the Redis set).
+      const redisValidate = await checkInvalidPasswordResetToken(token);
+      if (redisValidate) {
+        return response.status(400).json({
+          status: "fail",
+          message: "This password reset token is expired/invalid!",
+        });
+      }
```
lxkedinh: This can be moved to an express validator inside a `.custom()` check.

packages/server/routes/api/auth.js
```diff
@@ -283,4 +296,168 @@ router.post("/token/revoke", async (request, response) => {
...
+            message: "Could not generate a token from the provided userID.",
+          },
+        });
+      }
+
+      // Use template literal to embed the token as a query inside the link (Note: this is a testing link since the app is not hosted yet.)
+      const link = `http://localhost:5050/reset-password?token=${passwordResetToken}`;
+
+      // Uncomment the line below when the app is offically launched.
+      // const link = `https://icebreaksea.com/account/reset-password?token=${passwordResetToken}`;
```
lxkedinh: Like I said in my other comment about duplicate code in your sending email functions, create the link within the `sendPasswordResetEmail()` function instead.

packages/server/validators/auth.js
```diff
@@ -0,0 +1,27 @@
...
+const passwordResetValidator = [
+  body("token", "Invalid token.")
+    .exists({ checkFalsy: true })
+    .withMessage("Token can't be null or empty.")
+    .trim(),
+
+  body("password", "Invalid password.")
+    .exists({ checkFalsy: true })
+    .withMessage("Password can't be null or empty.")
+    .trim(),
```
lxkedinh: The order of your validator checks matter. Trim first and then check truthiness. The way you have it right now, a string with just whitespaces `"        "` will pass because it's a truthy value and then you're trimming it after checking truthiness.

lxkedinh (changes requested):

packages/server/utils/token.js
larrylaa: The userID that is being passed into the generateResetPasswordToken function inside of the route is already destructured. See getUserIdByEmail function in users.js

packages/server/validators/auth.js
```diff
@@ -1,24 +1,49 @@
 const { body } = require("express-validator");
+const AuthController = require("../controllers/auth");
+const UserController = require("../controllers/users");
+const { checkInvalidPasswordResetToken } = require("../utils/redis");
 
 const forgotPasswordValidator = [
```
lxkedinh: `forgotPasswordValidator` isn't a bad name, but taking a closer look at this validator shows all it does is validate user email in the request body. Renaming this to `userEmailValidator` is clearer and more specific to what it actually does. Naming it this way is clear when we potentially reuse it in future API routes too.

```js
router.post(
  "/forgot-password",
  userEmailValidator,
  // ...
});

File: packages/server/validators/auth.js [hunk 11]
```diff
@@ -0,0 +1,52 @@
...
+      if (isUsedToken) {
+        throw new Error("This password reset token is expired/invalid!");
+      }
+    }),
+
+  body("password", "Invalid password.")
+    .trim()
+    .exists({ checkFalsy: true })
+    .withMessage("Password can't be null or empty."),
+];
```
lxkedinh: For the actual password reset, most organizations require users to input their new password twice to confirm it, we should add a `passwordConfirmation` request body field to enforce that as well.

Validation should fail if `password` and `passwordConfirmation` in the request body don't match.

packages/server/validators/auth.js
```diff
@@ -0,0 +1,52 @@
...
+const forgotPasswordValidator = [
+  body("email", "Invalid email.")
+    .trim()
+    .exists({ checkFalsy: true })
+    .withMessage("Email can't be null or empty.")
+    .isEmail()
+    .withMessage("Invalid email entered!")
+
+    // Check If User Exists in DB & If User is a Google OAuth Acc
+    .custom(async (email) => {
```
lxkedinh: When any of our express-validators have to query the database for validation, we should use a `.bail()` check to stop validation right before the custom validator that queries our database.

```js
body("email", "Invalid email.")
  // ...
  .isEmail()
  .withMessage("Invalid email entered!")
  .bail()    // stop validation here if any of the previous checks fail to prevent querying database
  .custom(async (email) => {
  // ...
})
```

If the checks prior to the custom validator fail, it's a waste of time to run the custom validator that queries our database because we already know validation failed.

Read more about `.bail()` in docs [here](https://express-validator.github.io/docs/api/validation-chain/#bail).

packages/server/validators/auth.js
```diff
@@ -0,0 +1,52 @@
...
+      const isUsedToken = await checkInvalidPasswordResetToken(token);
+      if (isUsedToken) {
+        throw new Error("This password reset token is expired/invalid!");
+      }
+    }),
+
+  body("password", "Invalid password.")
+    .trim()
+    .exists({ checkFalsy: true })
+    .withMessage("Password can't be null or empty."),
```
lxkedinh: We should enforce some rules for passwords. These are just my initial thoughts and we can polish this out:
- 12 to 20 characters in length
- At least 1 uppercase character
- At least 1 number
- At least 1 special character from these `!@#$%^&*`

You should use a regular expression in the express-validator to enforce this.

packages/server/controllers/auth.js
```diff
@@ -189,6 +190,84 @@ async function authenticate(request, response, next) {
...
+  return info;
+}
+
+async function sendPasswordResetEmail(address, passwordResetToken) {
+  const subject = "Icebreak: Password Reset Request";
+  const link = `http://localhost:5050/reset-password?token=${passwordResetToken}`;
+  const body = `Please click the following link to reset your password: ${link}`;
+
+  sendEmail(address, subject, body);
+}
```
lxkedinh: I believe this needs to be
```js
return sendEmail(address, subject, body);
```
since you're assigning the result of this function to a variable in the forgot passsword API route but this function doesn't return anything.

packages/server/controllers/auth.js
```diff
@@ -189,6 +190,84 @@ async function authenticate(request, response, next) {
...
+
+async function sendPasswordResetConfirmationEmail(address) {
+  const subject = "Icebreak: Password Reset";
+  const body =
+    "Your Icebreak password has been changed. If you did not request a password change, please immediately contact us at icebreak@cppicebreak.com.";
+
+  sendEmail(address, subject, body);
+}
```
lxkedinh: Same thing as earlier comment about returning `sendEmail`.
```js
return sendEmail(address, subject, body);
```

packages/server/utils/token.js
lxkedinh: If that's the case, change the parameter for `generateResetPasswordToken()` to better reflect that, `userId` is more clear than just `user`.
```js
function generateResetPasswordToken(userId) {
  return jwt.sign(
    {
      userId,
    },
    process.env.TOKEN_SECRET,
    {
      expiresIn: "15m",
    }
  );
}
```

lxkedinh (changes requested):

packages/server/controllers/auth.js
```diff
@@ -240,15 +240,15 @@ async function sendPasswordResetEmail(address, passwordResetToken) {
   const link = `http://localhost:5050/reset-password?token=${passwordResetToken}`;
```
lxkedinh: We can't hardcode the link to localhost because this would work just for testing only. We'll have to use an environment variable for the base URL that changes depending on whether it's the development environment or production.

I'll send more details in Discord.

lxkedinh (changes requested):

lxkedinh (approved):

PR #48, opened 2024-03-02
Fetch upcoming events of a guild route

BLiu456:
Implementation of an API route that gets the upcoming events of a guild.

### DIFF
--- a/packages/server/controllers/events.js
+++ b/packages/server/controllers/events.js
@@ -122,6 +122,20 @@ async function getEventAttendees(eventId) {
   return query;
 }
 
+async function getUpcomingEvents(currentDate, guildId) {
+  const upcomingEvents = await prisma.events.findMany({
+    where: {
+      guildId: guildId,
+      startDate: { gt: currentDate },
+    },
+    orderBy: {
+      startDate: "asc",
+    },
+  });
+
+  return upcomingEvents;
+}
+
 module.exports = {
   getEvent,
   getEvents,
@@ -131,4 +145,5 @@ module.exports = {
   updateEvent,
   createEvent,
   getEventAttendees,
+  getUpcomingEvents,
 };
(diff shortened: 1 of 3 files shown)


packages/server/controllers/events.js
```diff
@@ -122,6 +122,20 @@ async function getEventAttendees(eventId) {
   return query;
 }
 
+async function getUpcomingEvents(currentDate, guildId) {
+  const upcomingEvents = await prisma.events.findMany({
+    where: {
+      guildId: guildId,
+      startDate: { gt: currentDate },
+    },
```
lxkedinh: Use `gte` for "greater than or equal to" for the edge case that `currentDate` is exactly when the event starts.

packages/server/validators/events.js
```diff
@@ -195,8 +195,27 @@ const eventIdValidator = [
...
+    .withMessage("Not a UUID")
+    .bail()
+    .custom(async (value) => {
+      try {
+        await GuildController.getGuild(value); //Check if guild exists by getting it
+      } catch (error) {
+        throw new Error(`No guild exists with an ID of ${value}`);
+      }
+    }),
+];
```
lxkedinh: Use the `guildIdValidator` already defined in `validators/guilds.js`. We should try to reuse code when we can.

lxkedinh (changes requested):

packages/server/routes/api/events.js
```diff
@@ -280,4 +281,47 @@ router.get(
...
+      const upcoming = await EventController.getUpcomingEvents(
+        currDate,
+        guildId
+      );
+
+      response.status(200).json({
+        status: "success",
+        data: {
+          upcoming: upcoming,
+        },
```
lxkedinh: Try to be more specific with naming response data so it's more clear to the frontend.
```js
{
  status: "success",
  data: {
    upcomingEvents: upcoming
  }
}
```

lxkedinh (approved):
Everything looks good! One last change but it's very minor and I'll approve the PR now. You can squash and merge after you resolve my last comment.

PR #49, opened 2024-03-08
Event Overview Screen: Refactored

tdang2180:
**Features**
- Implemented React Context to pass data globally for both event and feed screens
- Removed parameters to pass event data as an array to overview screen


packages/app/components/EventCard/EventCardText.js
```diff
@@ -48,29 +64,25 @@ function EventCardText(props) {
...
   );
 }
 
 EventCardText.propTypes = {
-  description: PropTypes.string,
-  location: PropTypes.string,
-  timeBegin: PropTypes.string,
-  timeEnd: PropTypes.string,
-  title: PropTypes.string,
+  event: PropTypes.array,
```
MinT-Napkin: I am getting a Failed prop type: invalid prop 'event' of type 'object' supplied to 'EventCardText', expected 'array'

Make sure to handle any warnings you may see.

packages/app/components/EventCard/EventCardText.js
```diff
@@ -48,29 +64,25 @@ function EventCardText(props) {
...
       <Text
         {...(isTitlePressed
           ? { style: styles.eventTitlePressed }
           : { style: styles.eventTitle })}
         onPress={onTitlePress}>
-        {props.title}
+        {event.title}
       </Text>
-      <Text style={styles.smallText}>📌 {props.location}</Text>
+      <Text style={styles.smallText}>📌 {event.location}</Text>
```
MinT-Napkin: A little thing you could do here, although i'd say it's optional, is if location and/or description is null,
hide their hide their text views or maybe set a default location/description such as N/A (former is probably better),
it makes the cards a little less empty when location and/or description are found null and takes up unnecessary space in the card.

MinT-Napkin (approved):
Lgtm! There's really only one thing you need to change, something to do with the event proptypes in EventCardText.js since that is giving warnings. The other thing is something I suggested, purely optional.

I'll approve so you can go ahead and merge this to main after your done with that!

packages/app/components/EventCard/EventCardText.js
tdang2180: 👍 Changed prop type to any

packages/app/components/EventCard/EventCardText.js
tdang2180: I also added this feature for start and end times.

PR #50, opened 2024-03-14
Event attendee status route

MichaelWuhu:


packages/server/controllers/events.js
```diff
@@ -121,6 +121,29 @@ async function getEventAttendees(eventId) {
...
 }
 
+async function updateEventAttendeeStatus(userId, eventId, status) {
+  // Validate status
+  const allowedStatusValues = ["NotInterested", "Interested", "Attending"];
+  if (!allowedStatusValues.includes(status)) {
+    throw new Error(
+      "Invalid status value. Allowed values are: NotInterested, Interested, Attending"
+    );
+  }
```
lxkedinh: You should use an express-validator to validate the `status` request body parameter instead of doing it here. It's best to keep functions small so that they only have one responsibility. It's better to keep code small, modular, and delegate responsibilities. `updateEventAttendeeStatus` will only worry about executing the Prisma query to update the status while the express-validator will handle validation of the `status` request body input parameter.

packages/server/controllers/events.js
```diff
@@ -121,6 +121,29 @@ async function getEventAttendees(eventId) {
...
+    where: {
+      userId_eventId: {
+        userId: userId,
+        eventId: eventId,
+      },
+    },
+    data: {
+      status: status,
+    },
+  });
```
lxkedinh: You should use [upsert](https://www.prisma.io/docs/orm/prisma-client/queries/crud#update-or-create-records) instead of update because this will fail if there is no existing record yet when trying to update.

packages/server/routes/api/events.js
```diff
@@ -281,6 +281,45 @@ router.get(
...
+    if (!result.isEmpty()) {
+      response.status(400).json({
+        status: "fail",
+        data: result.array(),
+      });
+      return;
+    }
+
+    const { userId, status } = request.body;
+    const { eventId } = request.params;
```
lxkedinh: Remember to use `matchedData(request)` to access the validated input data now instead of accessing `request.body` or `request.params` because we're using express-validator.

See [here](https://express-validator.github.io/docs/guides/getting-started#accessing-validated-data).

packages/server/routes/api/events.js
```diff
@@ -281,6 +281,45 @@ router.get(
...
+      const updatedEventAttendeeStatus =
+        await EventController.updateEventAttendeeStatus(
+          userId,
+          eventId,
+          status
+        );
+
+      response.status(200).json({
+        status: "success",
+        data: { updatedEventAttendeeStatus: updatedEventAttendeeStatus },
```
lxkedinh: `updatedEventAttendeeStatus` is a bit too long of a name for the response data. The data we're responding to the user with here is the event attendee info so just `eventAttendee` is a shorter but specific enough name to be clear to the frontend.

```js
data: {
  eventAttendee: updatedEventAttendeeStatus
}
```

lxkedinh (changes requested):

packages/server/validators/events.js
```diff
@@ -193,8 +194,41 @@ const eventIdValidator = [
...
+    .withMessage("status cannot be null or empty")
+    .custom(async (status) => {
+      // Validate status
+      const allowedStatusValues = ["NotInterested", "Interested", "Attending"];
+      if (!allowedStatusValues.includes(status)) {
+        throw new Error(
+          "Invalid status value. Allowed values are: NotInterested, Interested, Attending"
+        );
+      }
+    }),
```
lxkedinh: Don't use a custom validator for this since all you're doing is matching characters which is better suited for regular expressions. Express-validator comes with `.matches()` to validate against a regular expression. See docs [here](https://express-validator.github.io/docs/api/validation-chain/#matches). You can also look at the other validators for examples of regular expression checks that we've already used.

packages/server/validators/events.js
```diff
@@ -193,8 +194,41 @@ const eventIdValidator = [
...
+    .blacklist("<>")
+    .isUUID()
+    .withMessage("Invalid user ID provided")
+    .custom(async (value) => {
+      try {
+        await UserController.getUser(value);
+      } catch (error) {
+        throw new Error("User with ID ${value} not found");
+      }
+    }),
```
lxkedinh: Move this out into a separate validator like `userIdBodyValidator` in `validators/users.js` instead since it keeps the code modular, and it will be easier to reuse for future features. Justin's task is also using a userId body validator so it'll be easier to resolve the merge conflict later.

lxkedinh (changes requested):

packages/server/validators/users.js
```diff
@@ -18,14 +18,14 @@ const userIdValidator = [
...
 
+// validate userId in body instead of params
 const userIdBodyValidator = [
   body("userId")
     .notEmpty()
     .withMessage("User ID cannot be empty")
     .blacklist("<>")
     .isUUID()
     .withMessage("Invalid user ID provided")
-    .bail()
```
lxkedinh: Keep the `.bail()` check that was here previously. It's used to end validation earlier if prior validation fails to prevent unnecessarily running the following custom validator that runs a Prisma query.

packages/server/validators/events.js
lxkedinh, line 1: We also need to enforce updating attendee status only if the users are a member of the guild that is hosting the event. Modify `attendeeStatusValidator` to validate that the user is a part of the guild hosting the event. You can refer to `checkInTimeValidator` if you're stuck since Justin also had to do that for his task.

lxkedinh (changes requested):

packages/server/validators/events.js
```diff
@@ -203,7 +203,18 @@ const attendeeStatusValidator = [
...
+      const userId = req.body.userId;
+      const event = await EventController.getEvent(eventId);
+      const isMember = await GuildController.isGuildMember(
+        event.guildId,
+        userId
+      );
+      if (!isMember) {
+        throw new Error("User is not a member of the guild.");
+      }
+    }),
```
lxkedinh: This is the right idea but the value inside the async function parameter is not `eventId` in this case, it's `status`. The value that's passed to the handler function inside a custom validator is the field that the custom validator is attached to, `body("status")` in this case.

`eventId` can be accessed by `req.params.eventId` inside the async function since the `eventIdValidator` is ran before this in the route handler function in `routes/api/events.js` so `eventId is already validated.

It would be better to attach this custom validator to the `userId` body field instead. `status` isn't related to this validation check, so it doesn't make sense to attach this validator to the `status` body field.

```js
// attached to body("userId")
.custom(async (userId, { req }) => {
    const eventId = req.params.eventId;    // access eventId here
    const event = await EventController.getEvent(eventId);
    const isMember = await GuildController.isGuildMember(
      event.guildId,
      userId
    );
    if (!isMember) {
      throw new Error("User is not a member of the guild.");
    }
  })
```

See docs [here](https://express-validator.github.io/docs/guides/customizing#implementing-a-custom-validator) for custom validators.

lxkedinh (changes requested):

lxkedinh (approved):
Everything looks good to me, nice work guys!

PR #52, opened 2024-03-28
Guild's Archived Events API Route

BLiu456:
Implementation of the guild's archived events API route. Where the route will get events from a guild that have happened in the past month of when the request is made.


packages/server/controllers/events.js
```diff
@@ -135,6 +135,17 @@ async function getUpcomingEvents(currentDate, guildId) {
   return upcomingEvents;
 }
 
+async function getArchivedEvents(currDate, pastDate, guildId) {
+  const archivedEvents = await prisma.events.findMany({
+    where: {
+      guildId: guildId,
+      AND: [{ startDate: { gte: pastDate } }, { endDate: { lte: currDate } }],
+    },
+  });
```
lxkedinh: Let's also sort the archived events from most recent to least recent, so the events that happened most recently to the oldest events.

lxkedinh (changes requested):

lxkedinh (approved):

PR #53, opened 2024-04-04
Guild Member Management API Routes

mikephu:
Implemented API routes for managing guild members. These routes include a POST for adding members to an existing guild, a DELETE for removing members from said guild, and a PUT for managing the roles of its members. Access to these routes are reserved for those with specific roles within a guild.


packages/server/routes/api/guilds.js
```diff
@@ -239,7 +243,244 @@ router.get(
...
+        ) {
+          return response.status(403).json({
+            status: "fail",
+            data: {
+              userId:
+                "Access denied. Only authorized members may add users to an invite-only guild.",
+            },
+          });
+        }
+      }
```
lxkedinh: We're no longer using if statement chains for validation checks now that we have `express-validator`. These are part of input validation logic so move these to an express-validator.

That way, it keeps the code more maintainable by delegating responsibilities to certain parts of the code and each part of the code is only responsible for one single thing.
- Validators only do validation logic on input parameters.
- Route handler functions only deal with communication to and from the frontend.
- Controller functions do the actual operations to fulfill the purpose of the API route.

packages/server/routes/api/guilds.js
```diff
@@ -239,7 +243,244 @@ router.get(
...
+
+      if (client.userId === userId) {
+        return response.status(403).json({
+          status: "fail",
+          data: {
+            userId: "The guild owner cannot edit their own role.",
+          },
+        });
+      }
```
lxkedinh: Same thing in my other comment about moving this validation logic to an `express-validator` instead.

packages/server/routes/api/guilds.js
```diff
@@ -239,7 +243,244 @@ router.get(
...
+    try {
+      // If member does not exist, fail
+      if (!(await GuildController.isGuildMember(guildId, userId))) {
+        return response.status(403).json({
+          status: "fail",
+          data: {
+            userId: "Guild member does not exist.",
+          },
+        });
+      }
```
lxkedinh: Same thing in my other comment about moving this validation logic to an `express-validator` instead.

packages/server/routes/api/guilds.js
```diff
@@ -239,7 +243,244 @@ router.get(
...
+        userData.role >= clientData.role
+      ) {
+        return response.status(403).json({
+          status: "fail",
+          data: {
+            userId:
+              "Access denied. User is not authorized to perform this function.",
+          },
+        });
+      }
```
lxkedinh: Same thing in my other comment about moving this validation logic to an `express-validator` instead.

packages/server/controllers/guilds.js
```diff
@@ -60,7 +61,65 @@ async function guildExists(guildId) {
...
+          return [[key, value]];
+        }
+      })
+      .reduce((acc, [key, value]) => {
+        acc[key] = value;
+        return acc;
+      }, {});
+
+    return flattenedData;
+  }
```
lxkedinh: Remember to replace this with a better data flattener since you say this is temporary for now.

lxkedinh (changes requested):

packages/server/routes/api/guilds.js
```diff
@@ -351,47 +306,6 @@ router.put(
     const role = request.body.role;
```
lxkedinh:
```js
const role = request.body.role;
```
The role should also be taken from matchedRequest. Express-validator validates both route URL parameters and the request body.
```js
const role = data.role;
```

packages/server/validators/guilds.js
```diff
@@ -223,6 +224,72 @@ const updateGuildValidator = [
     .withMessage("Invite mode must be set to true or false."),
 ];
 
+const addGuildMemberValidator = [
+  param("guildId", "userId").custom(async (_, { req }) => {
```
lxkedinh:
```js
param("guildId", "userId")
```

I'm assuming you want to apply the same custom validator to both of these fields. This isn't the right syntax for this, it should be an array since only the first argument to this function is for the fields that you want to validate.
```js
param(["guildId", "userId"])
```
This can be seen from looking at the function definition.
image

The first argument to `param()` is the field(s) and the second argument is for the default error message for this field.

packages/server/routes/api/guilds.js
```diff
@@ -239,7 +243,126 @@ router.get(
...
+        status: "error",
+        message: error.message,
+      });
+    }
+  },
+);
+
+// Update guild member's role
+router.put(
+  "/:guildId/members/:userId/roles",
```
lxkedinh: Guild members can only have one role, and this route only deals with updating a singular guild member's role here, so it'd be better to keep the route singular instead of plural as well.
```
/:guildId/members/:userId/role
```

packages/server/validators/guilds.js
```diff
@@ -231,11 +298,39 @@ const updateGuildMemberRoleValidator = [
     .blacklist("<>")
     .matches(/^(?:Member|Officer|Owner)$/)
     .withMessage("Invalid role. Allowed values are: Member, Officer, or Owner"),
+  param("guildId", "userId").custom(async (_, { req }) => {
```
lxkedinh:
```js
param("guildId", "userId").custom(async (_, { req }) => {
  // ...
});
```

Same thing as my other comment about defining an express-validator for multiple fields. These should be in an array as the first argument.
```js
param(["guildId", "userId"])
```

lxkedinh (changes requested):

lxkedinh (approved):

PR #54, opened 2024-04-04
(feat): QR Code Generation for Event Check-in

JustinNgxyen:
API route that generates a QR code for event check-ins.

1. Guild owner/officer presses "Generate QR Code" button on event overview screen to create an image containing the generated QR code. Generating an image is more convenient so it could be sent to the event host's laptop projecting screen.
2. Members who want to attend the event scan the QR code with their phone.
3. Scanning the QR code opens specific screen on Icebreak app and calls check-in API route, passing the eventId  and userId.

[RFC](https://doc.clickup.com/9011073139/d/h/8chm43k-3771/c6121ab443aaf1d)


packages/server/controllers/events.js
```diff
@@ -206,6 +207,17 @@ async function addCheckInPoints(eventId, userId) {
...
+async function generateCheckInQRCode(eventId) {
+  const text = `icebreak://qr-code-check-in?eventId=${eventId}`;
+
+  try {
+    const qrCodeDataURL = await QRCode.toDataURL(text);
+    return qrCodeDataURL;
+  } catch (err) {
+    console.error(err);
+  }
+}
```
lxkedinh: Remember that our controller functions don't have try/catch. Any errors will be caught by the try/catch block inside the corresponding route handler function.

packages/server/routes/api/events.js
```diff
@@ -403,7 +402,44 @@ router.put(
...
+
+    try {
+      const generatedQRURI =
+        await EventController.generateCheckInQRCode(eventId);
+
+      response.status(200).json({
+        status: "success",
+        data: {
+          uri: generatedQRURI,
+        },
```
lxkedinh: Try to keep your variable and property names more specific and clear.
```js
data: {
  qrCodeURI: generatedQRURI
}
```

lxkedinh (changes requested):

lxkedinh (approved):

PR #58, opened 2024-04-05
Fetch Public Upcoming Events

larrylaa:
- Implemented paginated GET route for fetching the public upcoming events (determined by IsInviteOnly guild field)
- Added Express Validator checks for filtering the route's inputs
- Wrote unit tests

### DIFF
--- a/packages/server/__tests__/controllers/events.test.js
+++ b/packages/server/__tests__/controllers/events.test.js
@@ -1,0 +1,67 @@
+const { prismaMock } = require("../prisma_mock");
+const { getPublicUpcomingEvents } = require("../../controllers/events");
+
+describe("getPublicUpcomingEvents Unit Tests", () => {
+  const testEvents = [
+    {
+      eventId: "eventId1",
+      name: "Event 1",
+      startDate: new Date("2024-04-10"),
+      guilds: {
+        isInviteOnly: false,
+      },
+    },
+    {
+      eventId: "eventId2",
+      name: "Event 2",
+      startDate: new Date("2024-04-15"),
+      guilds: {
+        isInviteOnly: false,
+      },
+    },
+  ];
+
+  test("should fetch public upcoming events when action is 'next'", async () => {
+    const limit = 10;
+    const action = "next";
+    const eventId = "mockEventID";
+
+    prismaMock.events.findMany.mockResolvedValue(testEvents);
+
+    const result = await getPublicUpcomingEvents(limit, action, eventId);
+
+    expect(prismaMock.events.findMany).toHaveBeenCalledWith({
+      take: limit,
+      skip: 1,
+      cursor: { eventId: eventId },
+      orderBy: { eventId: "asc" },
+      where: {
+        guilds: { isInviteOnly: false },
+        startDate: { gt: expect.any(Date) },
+      },
+    });
+
+    expect(result).toEqual(testEvents);
+  });
+
+  test("should fetch public upcoming events when no action is provided", async () => {
+    const limit = 10;
+    const action = undefined;
+    const eventId = undefined;
+
+    prismaMock.events.findMany.mockResolvedValue(testEvents);
+
+    const result = await getPublicUpcomingEvents(limit, action, eventId);
+
+    expect(prismaMock.events.findMany).toHaveBeenCalledWith({
+      take: limit,
+      orderBy: { eventId: "asc" },
+      where: {
+        guilds: { isInviteOnly: false },
+        startDate: { gt: expect.any(Date) },
+      },
+    });
+
+    expect(result).toEqual(testEvents);
+  });
+});
--- a/packages/server/__tests__/controllers/events.test.js
+++ b/packages/server/__tests__/controllers/events.test.js
@@ -23,18 +23,17 @@ describe("getPublicUpcomingEvents Unit Tests", () => {
 
   test("should fetch public upcoming events when action is 'next'", async () => {
     const limit = 10;
-    const action = "next";
     const eventId = "mockEventID";
 
     prismaMock.events.findMany.mockResolvedValue(testEvents);
 
-    const result = await getPublicUpcomingEvents(limit, action, eventId);
+    const result = await getPublicUpcomingEvents(limit, eventId);
 
     expect(prismaMock.events.findMany).toHaveBeenCalledWith({
       take: limit,
       skip: 1,
       cursor: { eventId: eventId },
-      orderBy: { eventId: "asc" },
+      orderBy: { startDate: "asc" },
       where: {
         guilds: { isInviteOnly: false },
         startDate: { gt: expect.any(Date) },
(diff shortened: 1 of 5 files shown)


packages/server/controllers/events.js
```diff
@@ -47,6 +47,52 @@ async function getEvents(limit, action, eventId) {
...
+    case "next":
+      events = await prisma.events.findMany({
+        take: limit,
+        skip: 1,
+        cursor: {
+          eventId: eventId,
+        },
+        orderBy: {
+          eventId: "asc",
+        },
```
lxkedinh: Order by ascending start date instead of `eventId` since we want to display the soonest public upcoming events that are starting first. If we order by eventId, the events on the app would be out of order for their start times.

packages/server/controllers/events.js
```diff
@@ -47,6 +47,52 @@ async function getEvents(limit, action, eventId) {
...
+      });
+      break;
+    // first request made for first page, cursor is null
+    default:
+      events = await prisma.events.findMany({
+        take: limit,
+        orderBy: {
+          eventId: "asc",
+        },
```
lxkedinh: Same thing as my other comment about ordering by ascending start date, not `eventId`.

packages/server/validators/events.js
```diff
@@ -244,10 +244,36 @@ const checkInTimeValidator = [
...
+        value === ":cursor" ||
+        /^([A-Za-z0-9+/]{4})*([A-Za-z0-9+/]{4}|[A-Za-z0-9+/]{3}=|[A-Za-z0-9+/]{2}==)$/.test(
+          value,
+        )
+      ) {
+        return true; // Return true if the value is either ':cursor' or a valid Base64 string
+      }
+      return false; // Return false if the value is not ':cursor' and not a valid Base64 string
+    })
+    .withMessage("Cursor must be BASE64 Encoded!"),
```
lxkedinh: Referring to your comments here, if this is the case, `cursor` would be better suited as a request body parameter instead. That would fix the issues and you can use `.optional()` and `isBase64()`.

I'll update the RFC accordingly.

packages/server/routes/api/events.js
```diff
@@ -101,6 +103,88 @@ router.get(
...
+    const cleansedCursor = data.cursor;
+
+    try {
+      const eventLimit = parseInt(requestedEventLimit) || DEFAULT_EVENT_LIMIT;
+      // base64 decode cursor parameter if not null
+      const requestCursor = cleansedCursor
+        ? Buffer.from(cleansedCursor, "base64").toString()
+        : "";
+      let [currentPage, action, eventId] = requestCursor.split("___");
+      currentPage = parseInt(currentPage) || 1;
```
lxkedinh: This route is slightly different from the other get events route. This one will only go forward compared to the other one which allows for going to the previous and next pages.

The `cursor` parameter in this route is **only** the eventId that acts as the cursor for the starting point of the next page. It is not the same as the other route where it is `currentPage___(prev or next)___eventId`. You don't have to keep track of `currentPage` or `action` because we are only going forward. Only the `eventId` cursor is needed.

packages/server/routes/api/events.js
```diff
@@ -101,6 +103,88 @@ router.get(
...
+          },
+        });
+      }
+
+      // generate next cursors for infinite scrolling
+      const lastEventId = events[events.length - 1].eventId;
+
+      const nextCursor = Buffer.from(
+        `${currentPage + 1}___next___${lastEventId}`,
+      ).toString("base64");
```
lxkedinh: The `next` cursor is only the base64-encoded `eventId` of the last fetched event.

packages/server/routes/api/events.js
```diff
@@ -101,6 +103,88 @@ router.get(
...
+        response.status(200).json({
+          status: "success",
+          data: {
+            events: events,
+            cursor: {
+              next: nextCursor,
+            },
+          },
+        });
+      }
```
lxkedinh: There's no difference between the code in this if/else block. You can remove this since the response structure is the same.

packages/server/routes/api/events.js
```diff
@@ -101,6 +103,88 @@ router.get(
...
+
+    const requestedEventLimit = data.limit;
+    const cleansedCursor = data.cursor;
+
+    try {
+      const eventLimit = parseInt(requestedEventLimit) || DEFAULT_EVENT_LIMIT;
+      // base64 decode cursor parameter if not null
+      const requestCursor = cleansedCursor
+        ? Buffer.from(cleansedCursor, "base64").toString()
+        : "";
```
lxkedinh: Instead of decoding from base64 here in the route handler function, you can move this to a sanitizer in the express validator instead. Refer [here](https://express-validator.github.io/docs/guides/customizing/#implementing-a-custom-sanitizer).

packages/server/controllers/events.js
```diff
@@ -47,6 +47,52 @@ async function getEvents(limit, action, eventId) {
   return events;
 }
 
+async function getPublicUpcomingEvents(limit, action, eventId) {
+  let events;
+
+  switch (action) {
+    case "next":
```
lxkedinh: Using an `action` parameter and a switch statement doesn't make sense in this case because we only have one "action", going forward for infinite scrolling, unlike the other route.

Instead, you can keep the Prisma queries but replace the switch statement and test if the cursor `eventId` is null.

packages/server/controllers/events.js
```diff
@@ -47,6 +47,52 @@ async function getEvents(limit, action, eventId) {
   return events;
 }
 
+async function getPublicUpcomingEvents(limit, action, eventId) {
+  let events;
```
lxkedinh: Like my earlier comment about `action` being unnecessary because we're only going forward, the parameters can be reduced to just
```js
getPublicUpcomingEvents(limit, eventId)
```

packages/server/__tests__/controllers/events.test.js
lxkedinh, line 1: Modify unit tests as needed from the requested changes in my other comments.

lxkedinh (changes requested):

packages/server/validators/events.js
```diff
@@ -252,21 +252,14 @@ const fetchPublicUpcomingEventsValidator = [
...
-        return true; // Return true if the value is either ':cursor' or a valid Base64 string
-      }
-      return false; // Return false if the value is not ':cursor' and not a valid Base64 string
-    })
-    .withMessage("Cursor must be BASE64 Encoded!"),
+    .isBase64()
+    .withMessage("Cursor must be BASE64 Encoded!")
+    .customSanitizer((value) =>
+      value ? Buffer.from(value, "base64").toString() : " ",
+    ),
```
lxkedinh: We need to make sure that not any random Base64 strings are passed in but that they are Base64-encoded UUIDs. Include a `.isUUID()` check after the cursor value is sanitized.

lxkedinh (approved):

PR #60, opened 2024-04-19
Fetch User's Upcoming Interested/Attending Events

jimmywin656:
Implemented a GET route that returns all of a user's upcoming events only if they marked them as "Interested" or "Attending". To implement the route, I created a new controller function called getUserEvents and brought in the userIdValidator to validate the userId.

### DIFF
--- a/packages/server/controllers/events.js
+++ b/packages/server/controllers/events.js
@@ -212,17 +212,21 @@ async function getUserEvents(currentDate, userId) {
       eventAttendees: {
         some: {
           userId: userId,
-          status: {
-            in: ["Interested", "Attending"],
-          },
+          status: { in: ["Interested", "Attending"] },
         },
       },
       startDate: { gte: currentDate },
     },
-    orderBy: {
-      startDate: "asc",
+    select: {
+      eventId: true,
+      startDate: true,
+      eventAttendees: {
+        select: { status: true },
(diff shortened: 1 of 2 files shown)


packages/server/controllers/events.js
```diff
@@ -206,6 +206,30 @@ async function addCheckInPoints(eventId, userId) {
...
+    },
+    select: {
+      eventId: true,
+      startDate: true,
+      eventAttendees: {
+        select: { status: true },
+        where: { userId: userId },
+      },
+    },
+  });
```
lxkedinh, line 228: These will be displayed as a list of event info cards for the user, let's sort them by ascending start date for convenience.

packages/server/routes/api/events.js
```diff
@@ -403,7 +405,45 @@ router.put(
         message: error.message,
       });
     }
-  }
+  },
+);
+
+router.get(
+  "/:userId/events",
```
lxkedinh, line 412: After looking into it a bit more, I think it makes more sense to move this to the user router instead so it becomes `GET /users/:userId/events`. This follows REST API convention more since we're interested in the `events` (a subcollection) of a particular user with `userId` in the `users` collection.

lxkedinh (changes requested):

PR #62, opened 2024-04-29
New User Onboarding API Route

larrylaa:
- Added route for updating fields asked when onboarding a new user.
- Implemented input validation/filtering
- Wrote unit tests


packages/server/controllers/users.js
```diff
@@ -62,11 +62,27 @@ async function getGuildsForUser(userId) {
   return guilds;
 }
 
+async function updateNewUser(userId, userData) {
+  // convert age field into a int, because Int? in prisma schema allows an string age such as "25".
+  userData.age = parseInt(userData.age);
```
lxkedinh: Use a `.toInt()` sanitizer on the `express-validator` for the `age` input field instead. Sanitization of input data is part of the responsibility of our express-validators, not our controller functions.

https://express-validator.github.io/docs/api/validation-chain/#toint

packages/server/routes/api/users.js
```diff
@@ -89,7 +92,55 @@ router.get(
...
+      // check if user is new
+      const user = await UserController.getUser(userId);
+      if (user.isNew === false) {
+        return response.status(400).json({
+          status: "fail",
+          data: {
+            userId: "User is not new!",
+          },
+        });
+      }
```
lxkedinh: This is part of the validation process and should be moved to an express-validator instead of here in the route handler function.

packages/server/validators/users.js
```diff
@@ -50,7 +51,85 @@ const userIdBodyValidator = [
...
+  // Avatar Check: exist, be between 1 and 255 chars, and is URL!
+  body("avatar", "Invalid avatar.")
+    .exists({ checkFalsy: true })
+    .withMessage("avatar can't be null or empty.")
+    .blacklist("<>")
+    .trim()
+    .isLength({ min: 1, max: 255 })
+    .withMessage("Name length must be between 1 to 255 characters.")
+    .isURL()
+    .withMessage("Must be an URL"),
```
lxkedinh: For the avatar URL, it has to be specifically a URL to an image so add a `.matches()` check after this to check against a regular expression to make sure it ends in:
- `.jpg`
- `.png`

lxkedinh (changes requested):

packages/server/validators/users.js
```diff
@@ -50,7 +51,113 @@ const userIdBodyValidator = [
...
+    }),
+
+  // Interests Check: Must be an array.
+  body("interests", "Invalid interests.")
+    .blacklist("<>")
+    .trim()
+    .isArray()
+    .withMessage("Must be an array.")
+    .exists({ checkFalsy: true })
+    .withMessage("interests can't be null or empty."),
```
lxkedinh: We should be more strict and validate that this is:
- an array that only contains words (strings with only alphabet characters). You can use that with express-validator [wildcards](https://express-validator.github.io/docs/guides/field-selection/#wildcards)
- not an empty array by passing in length parameters in the `isArray()` check, see [here](https://express-validator.github.io/docs/api/validation-chain/#isarray)

lxkedinh: We should be more strict and validate that this is an array that only contains words (strings with only alphabet characters). You can use that with express-validator [wildcards](https://express-validator.github.io/docs/guides/field-selection/#wildcards).

packages/server/validators/users.js
```diff
@@ -50,7 +51,113 @@ const userIdBodyValidator = [
...
+    ),
+
+  // Major Check: exist, be between 1 and 50 chars
+  body("major", "Invalid major.")
+    .exists({ checkFalsy: true })
+    .withMessage("major can't be null or empty.")
+    .blacklist("<>")
+    .trim()
+    .isLength({ min: 1, max: 25 })
+    .withMessage("Major length must be between 1 to 25 characters."),
```
lxkedinh: This is set to up to 100 characters in the Prisma schema, modify to fit 100 characters instead.
```
major          String?         @db.VarChar(100)
```

packages/server/validators/users.js
```diff
@@ -50,7 +51,113 @@ const userIdBodyValidator = [
...
+    }),
+
+  // First Name Check: exist, be between 1 and 50 chars
+  body("firstName", "Invalid firstName.")
+    .exists({ checkFalsy: true })
+    .withMessage("firstName can't be null or empty.")
+    .blacklist("<>")
+    .trim()
+    .isLength({ min: 1, max: 50 })
+    .withMessage("firstName length must be between 1 to 255 characters.")
```
lxkedinh: Correct this error message to tell the user to use between 1 and 50 characters.

packages/server/routes/api/users.js
```diff
@@ -89,7 +92,43 @@ router.get(
         message: error.message,
       });
     }
-  }
+  },
+);
+
+router.put(
+  "/:userId/new",
```
lxkedinh: I think we should change this to a POST request. While we are updating the user here, this is specifically for onboarding and a user can only onboard once. We're not allowing users to request this route again so it's not idempotent. [Idempotence in PUT vs POST](https://restfulapi.net/rest-put-vs-post/)

Let's also change this route to `POST /:userId/onboarding` to be more specific instead of just "new".

lxkedinh (changes requested):

lxkedinh (approved):

PR #64, opened 2024-05-01
Refactoring to Hash Stored User Emails

JustinNgxyen:
We will hash the email first with sha-256 before we store them in the database for security reasons.

### DIFF
--- a/packages/server/controllers/auth.js
+++ b/packages/server/controllers/auth.js
@@ -5,7 +5,7 @@ const token = require("../utils/token");
 const bcrypt = require("bcrypt");
 const client = new OAuth2Client(process.env.WEB_CLIENT_ID);
 const nodemailer = require("nodemailer");
-const UserController = require("../../controllers/users");
+const UserController = require("./users");
 
 const NAMESPACE = "7af17462-8078-4703-adda-be2143a4d93a";
 
@@ -194,16 +194,8 @@ async function authenticate(request, response, next) {
 }
 
 async function isUserEmail(email) {
-  const emailHash = await UserController.hashUserEmail(email);
-  const result = await prisma.user.findUnique({
-    where: {
-      email: emailHash,
-    },
-  });
-
-  if (result === null) return false;
-
-  return true;
+  const user = await UserController.getUserByEmail(email);
+  return !!user;
 }
 
 async function isGoogleAccount(userId) {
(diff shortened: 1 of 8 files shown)


packages/server/controllers/auth.js
```diff
@@ -13,6 +14,9 @@ async function create(accessToken, refreshToken, profile, callback) {
     let { sub, given_name, family_name, picture, email } = profile._json;
     picture = picture.replace("=s96-c", "");
     const googleUUID = uuidv5(sub, NAMESPACE);
+    const hash = crypto.createHash("sha256");
+    hash.update(email);
+    const emailHash = hash.digest("hex");
```
lxkedinh: It would be better to extract this code into a function that can be used whenever needed instead of copy-pasting duplicate code whenever we want to hash an email. Move this into a `hashUserEmail()` function like in the Users controller and substitute the places you're currently hashing emails to call this new function instead.

lxkedinh (changes requested):

packages/server/controllers/users.js
```diff
@@ -24,18 +25,20 @@ async function getUserEmail(userId) {
 }
 
 async function getUserByEmail(email) {
+  const emailHash = hashUserEmail(email);
```
lxkedinh: The `hashUserEmail()` function you wrote is a promise so remember to await these.

packages/server/controllers/users.js
```diff
@@ -24,18 +25,20 @@ async function getUserEmail(userId) {
...
     where: {
-      email: email,
+      email: emailHash,
     },
   });
   return query;
 }
 
 async function getUserIdByEmail(email) {
+  const emailHash = hashUserEmail(email);
```
lxkedinh: Remember to use await for promises.

packages/server/controllers/auth.js
```diff
@@ -5,6 +5,7 @@ const token = require("../utils/token");
 const bcrypt = require("bcrypt");
 const client = new OAuth2Client(process.env.WEB_CLIENT_ID);
 const nodemailer = require("nodemailer");
+const UserController = require("../../controllers/users");
```
lxkedinh: This module path is incorrect. The `UserController` is in the same directory as the `AuthController`, the current file, so the path should be
```js
const UserController = require("../../controllers/users");
```

lxkedinh (changes requested):
The checks to query whether a user has the input email for the login and registration routes fail. You have to refactor to check against the email hash, not the email itself.

